PR #2, opened 2016-08-17
Minor doc improvement- consistent the caplization of 'appdaemon'

linville:
Minor change to make consistent the usage of 'appdaemon' when referring to the repo, config and python script itself in the README. Otherwise, commands in the README when copy/pasted may fail on case sensitive file systems.


acockburn:
Thanks for catching those :)

Issue #4, opened 2016-08-18
Allow logging to stdout and stderr


jasonmhite:
I'm managing AppDaemon via a systemd unit on Debian 8. The easiest way to do this is to actually not use the daemonize option and instead to run in the foreground. Systemd then handles managing the process and its children and also automatically captures stderr and stdout to the system log.

Currently when running in the foreground, log messages go stderr and errors get put into a `RotatingFileHandler` controlled by the `errorfile` config setting. It would be very convenient if I could set it to dump regular messages to stdout and errors to stderr, so that I can just let systemd handle organizing the logs; I'm on a RasPi and want to avoid writing to the filesystem (and have systemd configured to save logs in memory), plus it's handy to be able to view the logs with messages and errors colorized properly with `systemctl status`. 

To this end, I have a pull request where I have made a couple small changes. Basically, if you are running in the foreground and you have `logfile = STDOUT` in the configuration, then standard log messages will go to stdout. Similarly, if you have `errorfile = STDERR` then errors go to stderr. Finally, if you have both set, then it also suppresses the log messages that mention when an error has been logged (since you're going to see it in the output).

I did this in a way that I think won't break anything. For one, if called with `--daemonize` these checks are bypassed. Also, the previous behavior is preserved and this only kicks in when passing STDOUT or STDERR.

Let me know if you would prefer that it is implemented differently and I can change it.

acockburn:
Thanks for the PR I'll take a look soon - as long as it doesn't break previous behavior (e.g. logs to the terminal when running without Daemonize I think it is a good addition, especially given the detailed usecase you gave me :)

acockburn:
PR Merged

PR #5, opened 2016-08-18
Redirect log messages

jasonmhite:
PR for issue #4

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,8 +58,8 @@ timezone = <timezone>
 
 - `ha_url` is a reference to your home assistant installation and must include the correct port number and scheme (`http://` or `https://` as appropriate)
 - `ha_key` should be set to your key if you have one, otherwise it can be removed.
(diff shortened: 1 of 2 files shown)


acockburn:
Thanks - it would be nice if you could update the docs as well

jasonmhite:
OK, I added to `README.md`.

One question though, would you like me to add handling code to make calling `--daemonize`  with `logfile = STDOUT` or `logfile = STDERR` raise an exception? Current behavior is that it would write to a file named STDOUT and/or STDERR since I specifically did not mess with the `--daemonize` code path.

acockburn:
Yes, I think the exception is a good idea - thanks for the README changes.

jasonmhite:
Alright! I added a `ValueError` exception if you try and run in daemon mode with either redirect set in the config file.

Let me know if you want any other changes, otherwise I think it should be OK?

jasonmhite:
Also not related, but I really like this project!

acockburn:
Thanks a lot for your work, and glad that you like it :)

PR #12, opened 2016-08-29
Add select value to update input_sliders

djbanks:
Allow an input_slider to have its value updated from the app api.

### DIFF
--- a/API.md
+++ b/API.md
@@ -1133,6 +1133,36 @@ self.toggle("switch.patio_lights")
 self.toggle("light.office_1", color_name = "green")
 ```
 
+### select_value()
+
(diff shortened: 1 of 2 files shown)


acockburn:
Thanks for this - can you add an entry to the API doc as well please?

djbanks:
I have added an entry in the API doc. Hopefully it is acceptable.

acockburn:
Perfect thanks!

I am working on a new version - I still need a few days of testing as I have made some major changes to the scheduler but I will include this - hopefully you should see it next weekend.

acockburn:
Merged manually - will be in 1.3.1

PR #40, opened 2017-02-08
Custom Logger implementation

yawor:
I've created a custom logger implementation which allows inserting correct line numbers, filenames and function names using `Formatter` instead of string substitution. Whole implementation is in new `appdaemon.logging` module.

Classes in the new module:

- `CallStackManager` - A helper class which allows recording a call stack. It implements a context management protocol and keeps a record as a thread local attribute. It is instantiated in the module as a global variable call_stack_manager.
- `CallStackManagerLogger` - A custom Logger class which uses call_stack_manager to retrieve correct frame to log caller info (file, line no and func name).
- `CustomFormatter` - it overrides time formatting to work in the same way as the current logging logic.

There's also a `log_limit` decorator, which is used on `homeassistant.log` function and both logging methods on `AppDaemon` class. The decorator registers log calls on the CallStackManager.

I've set the default format string to:
`%(asctime)s %(levelname)s %(appname)s %(message)s`
where:

- `%(asctime)s`, `%(levelname)s` and `%(message)s` are standard logger fields (asctime formatting is implemented in `CustomFormatter`)
- `%(message)s` contains only the message (no app name, time and level - those are in their own fields).
- `%(appname)s` is the name of the app (as it was before in the log message itself). If app name is not provided, it defaults to "appdaemon".

Custom format string can be changed by setting `logformat` setting in `[AppDaemon]` section in the config file. To get caller info, following fields can be added (those are standard field in python logging):

- `%(filename)s` - contains name of the file where log function was called from
- `%(funcName)s` - contains calling function name where the log function was called
- `%(lineno)d` - contains line number where the log function was called

An example format string containing these fields:

`%(asctime)s %(levelname)s %(appname)s [%(filename)s - %(funcName)s@%(lineno)d] %(message)s`

PS. This change removes current string replacement. Message should no longer contain `__module__`, `__line__` and `__function__` placeholders as they won't be replaced anymore.

### DIFF
--- a/appdaemon/appapi.py
+++ b/appdaemon/appapi.py
@@ -7,6 +7,7 @@ import re
 import requests
 import inspect
 
+from appdaemon.logging import log_limit
 import appdaemon.homeassistant as ha
 
 
@@ -38,16 +39,6 @@ class AppDaemon:
         if service.find("/") == -1:
             raise ValueError("Invalid Service Name: {}".format(service))
 
-    def _sub_stack(self, msg):
-        stack = inspect.stack()
-        if msg.find("__module__") != -1:
-            msg = msg.replace("__module__", stack[2][1])
-        if msg.find("__line__") != -1:
-            msg = msg.replace("__line__", str(stack[2][2]))
-        if msg.find("__function__") != -1:
-            msg = msg.replace("__function__", stack[2][3])
-        return msg
-
     #
     # Utility
     #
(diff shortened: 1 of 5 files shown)


acockburn:
Sorry, I'm not going to merge this now - too much has changed in the underlying code. I also think that anyone who wants this kind of functionality is probably savvy enough to write it themselves, all I need to do is give them access to the logger objects, so I'll add calls for that in 3.0.

PR #42, opened 2017-02-09
Respect cert_path setting when connecting to WebSocket over SSL

yawor:
When analysing an issue #41 I've found out that the WebSocket connection doesn't respect cert_path setting for self signed certificates. This PR fixes this.

### DIFF
--- a/appdaemon/appapi.py
+++ b/appdaemon/appapi.py
@@ -267,10 +267,10 @@ class AppDaemon:
             headers = {'x-ha-access': conf.ha_key}
         else:
             headers = {}
-        apiurl = "{}/api/events/{}".format(
-            conf.ha_url, event, verify=conf.certpath
+        apiurl = "{}/api/events/{}".format(conf.ha_url, event)
+        r = requests.post(
+            apiurl, headers=headers, json=kwargs, verify=conf.certpath
         )
-        r = requests.post(apiurl, headers=headers, json=kwargs)
         r.raise_for_status()
         return r.json()
 
--- a/appdaemon/appdaemon.py
+++ b/appdaemon/appdaemon.py
@@ -1358,7 +1358,12 @@ def run():
                 elif url.startswith('http://'):
                     url = url.replace('http', 'ws', 1)
 
-                ws = create_connection("{}/api/websocket".format(url))
+                sslopt = {}
+                if conf.certpath:
+                    sslopt['ca_certs'] = conf.certpath
+                ws = create_connection(
+                    "{}/api/websocket".format(url), sslopt=sslopt
+                )
                 result = json.loads(ws.recv())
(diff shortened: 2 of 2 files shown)


joopert:
Is there anyone who can merge this PR? I ran into the same issue and wanted to do a PR as well, but found out there was a PR open already. @acockburn ?

acockburn:
I've been holding off on merging PRs because I have been been busy with HADasboard v2 which have been making a lot of changes to AppDaemon and I wanted to avoid any merge problems, I will look at this and the other PRs soon.

yawor:
@acockburn I don't want to pry as this is your project (and this issue is certainly not a place to discuss that, sorry), but wouldn't it be better to leave AppDaemon as is (I mean the purpose) and put the dashboard into a separate project?

Why make it heavier than it currently is? I wonder how many of current AppDaemon users is actually in need for such functionality. If it's about some integration between apps and dashboard widgets, then I think it's still possible to make this as a separate application/project. There's already a communication mechanism in form of HA event bus (you can send your own events for example from dashboard application to AppDaemon apps).

acockburn:
I did think about it in some depth and I had several reasons for doing it this way:

- I am aware of quite a few users that use both
- There are benefits to HA in this setup - AppDaemon caches HAs state and prevents the dashboard from launching "query storms" against HA when a page is loaded as the existing Dash does. Of course I could build that into the new dashboard but at that point I am duplicating significant amounts of code which I hate to do
- In the actual implementation, for non-dashboard users, the dashboard code is completely shutdown. Likewise for non-appdaemon users, thje app code, scheduler etc are all shutdown, so I don;t believe there is a significant increase in heaviness for users using just one or the other.
- Most of the dashboard code is isolated for easy maintenance
- I feel like it is easier to maintain one project than 2
- Communication between the Dash and AppDaemon has a lot of potential - why involve a 3rd party, greater load on HA and additional latency if it isn't necessary? I am very mindful of the fact that many many users are running on Pi 2 hardware which is quite slow - I like to avoid additional work where possible

For these reasons I decided to integrate the new Dashboard with AppDaemon. I did consider the alternative but in the end I decided on this route.

yawor:
@acockburn OK that's fair enough :).

@joopert If you really need the WS SSL fix right now, you can checkout my branch for now. If you've used git to install AppDaemon, go to the appdaemon dir and execute following commands:

- `git remote add yawor https://github.com/yawor/appdaemon.git`
- `git fetch yawor`
- `git checkout --track yawor/fix_ws_selfsigned_ssl`

First command adds new remote repository (named `yawor`). Second fetches new remote's metadata. Third checks out the branch, on which this PR is based.

acockburn:
@yawor Thanks for your comments in PR55 - I will make sure to include in this PR when I merge it.

PR #56, opened 2017-04-18
Some Docker Help

quadportnick:
- Changes to Dockerfile/.dockerignore
- Added a script for AD under Docker that streamlines setting things up
- Added a tutorial

Still need a little bit of help for the PiFolk, if it's just documentation or what have you. Also once it's in there lets play with having Docker built on an officially named repo and update the tutorial to point to that.


acockburn:
Thanks for this Nick, and sorry for the delay in responding, I am traveling on business. I'll plan to get this into the next beta.

quadportnick:
No problem :) Do you happen to know who's responsible for homeassistant on Docker Hub? I'd think the next step would be getting the auto-build linked up on something official, then we can remove the quadportnick refs from the guide and point there.

acockburn:
Not sure why I closed this :)

I probably won't merge as is, but I am now looking at adding hass.io support and improving basic docker support at the same time and this is very helpful!

PR #57, opened 2017-04-26
Fixed bug where media title was not displayed

mad-ady:
Tested with mpd. The browser would receive the media_title attribute from HA, but would not display it. It seems the code was copy pasted and the attribute name was not changed. I can confirm that the widget now displays the correct media name.

### DIFF
--- a/appdaemon/widgets/basemedia/basemedia.js
+++ b/appdaemon/widgets/basemedia/basemedia.js
@@ -159,7 +159,7 @@ function basemedia(widget_id, url, skin, parameters)
         {
             self.set_field(self, "album", state.attributes.media_album_name)
         }
-        if ("media_album_name" in state.attributes)
+        if ("media_title" in state.attributes)
         {
             if ("truncate_name" in self.parameters)
             {


homeassistant:
Hi @mad-ady, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;appdaemon#57).

Thanks!

acockburn:
Hi There - 

Thanks for the heads up - I think that this is an area that differs between media players, as what I have works fine for SONOS. I won't merge this PR directly but I'll use it as input to a review I have planned of the different media players out there and how they work as they all seem to be different.

mad-ady:
Ok, let me know if you need additional info to reproduce this issue (I used
mpd as a media player). The format should have been standardized inside HA,
but if it is not, then there should be a way to map HA fields into
dashboard fields. But I don't know how easy that would be.

mad-ady:
Hi! I've re-hit this bug and the fix is the same as the PR. Once you make the change to display media_title, it works great. I would like to see this fix in the main code (so that I don't have to re-add it after each update). I'm using Music Player Daemon 0.20.12 and I can send you the necessary HA and Dashboard configuration, together with a test radio stream to reproduce the issue if needed.

If you were to enable it - what would be the worst problem you expect to see? If media_title is not present in the attributes (not supported by the media_player), the code is skipped. If it is present here and somewhere else (media_artist, media_album_name) you might see the field twice.

Alternatively would you accept a PR with a new configuration option to display media_title that would enable this on demand?

Thanks.

PR #61, opened 2017-05-06
Added locale to clock

biacz:
users could change locale with a parameter in their panel configuration
i tried to add more options but i am not a coder and failed miserably. 
basically you could also format the whole date display using:
https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString

### DIFF
--- a/appdaemon/widgets/baseclock/baseclock.js
+++ b/appdaemon/widgets/baseclock/baseclock.js
@@ -48,7 +48,7 @@ function baseclock(widget_id, url, skin, parameters)
 		s = today.getSeconds();
 		m = formatTime(m);
 		
-        self.set_field(self, "date", today.toLocaleDateString());
+        self.set_field(self, "date", today.toLocaleDateString(self.parameters.locale));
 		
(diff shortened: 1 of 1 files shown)


biacz:
this is what i got to change the format of my clock for example:
function updateTime(self)
	{
		var today = new Date();
    var options = { weekday: 'long', year: 'numeric', month: 'long', day: 'numeric' };
    h = today.getHours();
		m = today.getMinutes();
		s = today.getSeconds();
		m = formatTime(m);

        self.set_field(self, "date", today.toLocaleDateString(self.parameters.locale, options));

		if ("time_format" in self.parameters && self.parameters.time_format == "24hr")
		{
			time = h + ":" + m;
			pm = ""
		}
		else
		{
			time = formatHours(h) + ":" + m;
			pm = " " + formatAmPm(h)
		}

		if ("show_seconds" in self.parameters && self.parameters.show_seconds == 1)
		{
			time = time + ":" + formatTime(s)
		}

		time = time + pm
		self.set_field(self, "time", time);
	}

PR #64, opened 2017-06-01
Add missing offset parameter

cgtobi:


acockburn:
Thanks for this - however I believe the docs as they stand are accurate, since offset is actually a keyword argument, and included under **kwargs - the docs make this clear. If you can show me an example of this being done differently in other python docs I'll be happy to revisit.

cgtobi:
I understand that, however it is inconsistent, see the following.

`self.handle = self.run_at_sunset(callback, offset, **kwargs)`
and
`self.handle = self.run_at_sunrise(callback, **kwargs)`

acockburn:
OK, yes I agree - I fixed sunset to match sunrise in the latest beta.

PR #69, opened 2017-06-22
Fix argument formatting in README

colinodell:
The argument section looks fine in a text editor but does not render nicely in Markdown.  By wrapping that content with ` ``` ` we can display the text with a monospace font and no inner Markdown rendering.

Visual diff:

image


acockburn:
Fixed in version 2.0.0 (not merged directly)

PR #84, opened 2017-07-14
Fix websocket to use wss when using https

jgrieger1:
When using https, websocket was still using ws.  the location.protocol includes the colon so the comparison failed and defaulted to using ws.


homeassistant:
Hi @jgrieger1, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;appdaemon#84).

Thanks!

acockburn:
Good catch - thanks!

PR #85, opened 2017-07-15
Added persistent_notification support.

fabianonline:
Hi.

I've added support for persistent_notifications to the current version of AppDaemon / HADashboard.

bildschirmfoto 2017-07-15 um 12 52 50

So far, this is working but pretty work-in-progress. But I had to make a few changes to the core of AppDaemon and would like to get a bit of feedback on this. If that feedback is positive, I will continue to work on polishing this PR. So, to be clear, I do not expect this Pull Request to be merged. (But if you want to, feel free to do it.)

What I changed in the core:
* Widgets can be "gridless" by setting "gridless: 1" in their YAML file. This means they won't take (or get) any space in the grid and will instead be placed at the bottom of `body` - initially hidden, in the most cases.
* Added a path `/sounds` to the router in order to be able to provide sound files for the browser to play.
* Widgets can have a callback `initial_with_null`. This behaves more or less like `initial` except that `initial_with_null` will be called with `null` if the entity doesn't exist at dashboard start. This may be the case with `persistent_notifications`, because they don't have to be declared before use.
* Callback for clicks on Widgets now support a parameter `live`, which will use jQuery's `.on` function for setting the `click` handler. This is necessary for attaching those events to hidden elements.
* A widget will not only have `entity` set, but also `short_entity`, which is the `entity` without the domain. (`persistent_notification.alert` will be `short_entity: alert`.)

To use / test:
* Checkout the code.
* Add a sound file as `/appdaemon/assets/sounds/tos-redalert.mp3`. Guess which sound file I used. ;-) For copyright reasons, I didn't include that file in that commit.
* Add a `persistent_notification` to the layout of a dashboard. For example, adding `persistent_notification.alert` somewhere. (Doesn't matter where in the output, it won't appear in the grid after all.)
* Fire a persistent_notification. For example by using "Services" in the "Developer Tool" of Home Assistant and then sending "Domain: persistent_notification", "Service: create" and "Service Data: `{"notification_id":"alert", "title":"Notification", "message":"This is a test notification."}`".

If you have any questions, feel free to ask me.

Oh, and, by the way, thanks for this great tool.

### DIFF
--- a/appdaemon/appdash.py
+++ b/appdaemon/appdash.py
@@ -37,6 +37,7 @@ def set_paths():
     conf.compiled_css_dir = os.path.join(conf.compile_dir, "css")
     conf.fonts_dir = os.path.join(conf.dash_dir, "assets", "fonts")
     conf.images_dir = os.path.join(conf.dash_dir, "assets", "images")
+    conf.sounds_dir = os.path.join(conf.dash_dir, "assets", "sounds")
     conf.base_url = "http://{}:{}".format(conf.dash_host, conf.dash_port)
     conf.stream_url = "ws://{}:{}/stream".format(conf.dash_host, conf.dash_port)
 
@@ -281,6 +282,8 @@ def setup_routes():
 
     # Add static path for images
     app.router.add_static('/images', conf.images_dir)
+
+    app.router.add_static('/sounds', conf.sounds_dir)
 
 
 # Setup
--- a/appdaemon/assets/javascript/dashboard.js
+++ b/appdaemon/assets/javascript/dashboard.js
@@ -123,6 +123,9 @@ var WidgetBase = function(widget_id, url, skin, parameters, monitored_entities, 
                             {
                                 child.ViewModel.title("entity not found: " + entity.entity);
                                 new_state = null
+                                if (entity.initial_with_null) {
+                                    entity.initial_with_null(child, new_state)
+                                }
                             }
                             else
                             {
@@ -198,15 +201,19 @@ var WidgetBase = function(widget_id, url, skin, parameters, monitored_entities, 
     clen = callbacks.length;
     for (i=0;i < clen;i++)
     {
-        $(callbacks[i].selector).click((
-            function(callback, ch, params)
+        var func = function(callback, ch, params)
+        {
+            return function()
             {
-                return function()
-                {
-                    callback(ch, params)
-                };
-            }(callbacks[i].callback, child,callbacks[i].parameters))
-        );
+                callback(ch, params)
+            };
+        }(callbacks[i].callback, child, callbacks[i].parameters);
+
+        if (callbacks[i].live) {
+            $(callbacks[i].selector[0]).on('click', callbacks[i].selector[1], func);
+        } else {
+            $(callbacks[i].selector).click(func);
+        }
     }
     
     // Create and initialize bindings
--- a/appdaemon/assets/templates/dashinit.jinja2
+++ b/appdaemon/assets/templates/dashinit.jinja2
@@ -14,8 +14,12 @@ $(function(){ //DOM Ready
     // Add Widgets
 
     var gridster = $(".gridster ul").gridster().data('gridster');
-    {% for widget in widgets%}
-        gridster.add_widget('<li><div data-bind="attr: {style: widget_style}" class="widget widget-{{widget.parameters.widget_type}}-{{widget.id}}" id="{{widget.id}}">{{widget.html}}</div></li>', {{widget.size[0]}}, {{widget.size[1]}}, {{widget.position[0]}}, {{widget.position[1]}})
+    {% for widget in widgets %}
+        {% if widget.parameters.gridless==1 %}
+            $('body').append('<div data-bind="visible: display, attr: {style: persistent_notification_style}" class="widget widget-{{widget.parameters.widget_type}} widget-{{widget.parameters.widget_type}}-{{widget.id}}" id="{{widget.id}}">{{widget.html}}</div>')
+        {% else %}
+            gridster.add_widget('<li><div data-bind="attr: {style: widget_style}" class="widget widget-{{widget.parameters.widget_type}}-{{widget.id}}" id="{{widget.id}}">{{widget.html}}</div></li>', {{widget.size[0]}}, {{widget.size[1]}}, {{widget.position[0]}}, {{widget.position[1]}})
+        {% endif %}
     {% endfor %}
     
     
--- a/appdaemon/dashboard.py
+++ b/appdaemon/dashboard.py
@@ -326,9 +326,9 @@ def add_layout(value, layout, occupied, dash, page, includes, css_vars, global_p
         while "{}x{}".format(column, layout) in occupied:
             column += 1
 
+        widget = {}
         if name != "spacer":
             sanitized_name = name.replace(".", "-").replace("_", "-").lower()
-            widget = {}
             widget["id"] = "{}-{}".format(page, sanitized_name)
 
             if widget_exists(dash["widgets"], widget["id"]):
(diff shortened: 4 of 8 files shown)


homeassistant:
Hi @fabianonline, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;appdaemon#85).

Thanks!

acockburn:
This is very interesting - and it dovetails nicely with some other work I have been planning, and could be a great basis, appreciate the work you have done to date. I think with some tweaks I'd be happy to merge this for Persistent Notifications and use it as a base for other things. Some more in-depth commentary inline:

* Widgets can be "gridless" by setting "gridless: 1" in their YAML file. This means they won't take (or get) any space in the grid and will instead be placed at the bottom of `body` - initially hidden, in the most cases.

[aimc] I think the gridless approach is interesting, but I forsee a lot of cases where a very similar dialog could be used, to the point where I think it could be a builtin to all dashboards without needing to be explicitly declared:

- Persistent Notifications (obviously)
- Keypad for the Alarm Widget
- Detailed popup for dimming lights and setting their color
- Temperature slider popup
- Many more

So either we would have one reusable dialog, or add something to the widget declaration that automatically adds a gridless dialog for any widget that needs it. Thoughts?

I'd also be interested if you can see any way of positioning this centrally without using `position: fixed;` - that seems to run afoul of some browser bugs and has caused trouble with the skinning when I used it for the Alarm keypad implementation.

* Added a path `/sounds` to the router in order to be able to provide sound files for the browser to play.

[aimc] Perfect

* Widgets can have a callback `initial_with_null`. This behaves more or less like `initial` except that `initial_with_null` will be called with `null` if the entity doesn't exist at dashboard start. This may be the case with `persistent_notifications`, because they don't have to be declared before use.

[aimc] I like that this is optional, but I would prefer the name to be a little more descriptive, perhaps "initial_entity_not_found"? This is actually a great idea to allow widgets to be a bit smarter than they currently are and have code to popup an alert instead of writing into the title, so I love the idea :)

Also, from reading the code, I think that the entity not found error should be suppressed if the initial_entity_not_found() is present - at the moment it looks like it does that anyway. Given my comment above, I/We will probably want to remove that code anyway as we can handle that case in the Widgets with this new framework.

* Callback for clicks on Widgets now support a parameter `live`, which will use jQuery's `.on` function for setting the `click` handler. This is necessary for attaching those events to hidden elements.

[aimc] I'll take your word for this, but would like a better explanation so I understand what the issue is - is it simply that the existing click handler doesn't work on hidden elements?

* A widget will not only have `entity` set, but also `short_entity`, which is the `entity` without the domain. (`persistent_notification.alert` will be `short_entity: alert`.)

[aimc] Again from reading the code, I'm not sure why we need this? Also, can't we derive the short name in the code rather than making the user add it as a parameter? I prefer consistency :)

Could you also, just so I have the whole picture show me what this looks like configured in a dashboard file?

Overall, I think this is really promising and I am looking forward to working with you some more on this!

fabianonline:
> [aimc] I think the gridless approach is interesting, but I forsee a lot of cases where a very similar dialog could be used, to the point where I think it could be a builtin to all dashboards without needing to be explicitly declared:
> 
> Persistent Notifications (obviously)
> Keypad for the Alarm Widget
> Detailed popup for dimming lights and setting their color
> Temperature slider popup
> Many more
> So either we would have one reusable dialog, or add something to the widget declaration that automatically adds a gridless dialog for any widget that needs it. Thoughts?

One reusable dialog could lead to annoying bugs when it gets "reused" by an incoming notification. But I think two dialogs could be enough. One for detailed popups and stuff and one "more important" one for notifications, which can then be above the detailed popup. On the other hand, allowing each widget to have as many dialogs as necessary should be the more future-proof solution.

Notifications working "out-of-the-box" would be a nice feature. Definitely more easy to use than my "add this widget to the layout wherever you want because it will not be showing up at all" approach.
I would love to be able to use wildcards or whatever... `persistent_notification.alert_*` being red and with sound, `persistent_notification.warning_*` being yellow and without sound... And normal notifications for stuff like displaying a box for 10 seconds reading "Ding Dong" when someone rings the doorbell for example...

> I'd also be interested if you can see any way of positioning this centrally without using position: fixed; - that seems to run afoul of some browser bugs and has caused trouble with the skinning when I used it for the Alarm keypad implementation.

Fixed positioning has the advantage of always being on screen and in the middle. Other positionings could lead to the notification being off-screen if you have large dashboards with scrolling. But I'm not a CSS pro - perhaps there are other ways to achieve something like this.

> Widgets can have a callback initial_with_null. This behaves more or less like initial except that initial_with_null will be called with null if the entity doesn't exist at dashboard start. This may be the case with persistent_notifications, because they don't have to be declared before use.
> [aimc] I like that this is optional, but I would prefer the name to be a little more descriptive, perhaps "initial_entity_not_found"? This is actually a great idea to allow widgets to be a bit smarter than they currently are and have code to popup an alert instead of writing into the title, so I love the idea :)

Yeah, that sounds reasonable.

> Callback for clicks on Widgets now support a parameter live, which will use jQuery's .on function for setting the click handler. This is necessary for attaching those events to hidden elements.
> [aimc] I'll take your word for this, but would like a better explanation so I understand what the issue is - is it simply that the existing click handler doesn't work on hidden elements?

You are calling `$(selector).click(function...)` to add an onClick handler on the HTML element matching the selector. But apparently (backed by a short Google search) it isn't possible to add an onClick handler to elements that are not visible. Since the notification is hidden by default, it is thereby impossible to add the onClick handler to the Dismiss button. Using `.on()` sets a "global" onClick handler that, when clicked, checks it the clicked element matches the selector.

> A widget will not only have entity set, but also short_entity, which is the entity without the domain. (persistent_notification.alert will be short_entity: alert.)
> [aimc] Again from reading the code, I'm not sure why we need this? Also, can't we derive the short name in the code rather than making the user add it as a parameter? I prefer consistency :)

Forget that. (At least the modification to the core.) Problem is: `persistent_notification.dismiss` doesn't accept an `entity_id` parameter to tell it which notification to dismiss, but instead expects a `notification_id` which is simply the `entity_id` with `persistent_notification.` removed.

So I put code in dashboard.py to add the `short_entity` field to the widget for using that when making the request to dismiss the notification.

But in the meantime I realized I can just make that modification to the entity_id in the JS code and don't have to modify the code. I'll remove that code later today / tomorrow.

> Could you also, just so I have the whole picture show me what this looks like configured in a dashboard file?

Well, it's really not much...

```
title: Wohnzimmer
widget_dimensions: [83, 74]
widget_margins: [5, 5]
widget_size: [2, 2]
columns: 12
global_parameters:
  use_comma: 1
  time_format: "24hr"
  show_seconds: 1
  precision: 1
  shorten: 0


wohnzimmer_deckenlampe:
  [snip]
wohnzimmer_fernsehlicht:
  [snip]
wohnzimmer_lichterkette:
  [snip]
wohnzimmer_level:
  [snip]
flash_wohnzimmer:
  [snip]
turn_off_wohnzimmer:
  [snip]

layout:
    - include: header
    - wohnzimmer_deckenlampe, light.wohnzimmer_stehlampe, wohnzimmer_lichterkette, wohnzimmer_fernsehlicht, wohnzimmer_level
    - persistent_notification.alert
    - turn_off_wohnzimmer, flash_wohnzimmer
    - empty: 3
    - include: footer
```

Only thing for the notifications is the widget in line 3 of the layout.

> Overall, I think this is really promising and I am looking forward to working with you some more on this!

Great to hear that.

acockburn:
OK, I think the way to go is this:

1. Persistent Notifications seem to be a special case for additional dialogs. I think the dialog should be added implicitly to every dashboard and should just transparently work with no config if possible
2. All other widgets that need dialogs will be explicitly configured anyway and can use the gridless concept, although again, there should be no need to explicitly declare the additional dialog it should just happen

Question - if we are removing the concept of explicitly declaring any of the dialogs, do we still need "gridless"?

PR #94, opened 2017-07-22
Dashboard: don't use absolute base_url

trnila:
When I run dashboard behind proxy, application does not work because it uses bind address that is not accessible. When we don't set absolute base_url, we don't need to set additional configuration to get it working anywhere.


homeassistant:
Hi @trnila, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;appdaemon#94).

Thanks!

acockburn:
Thanks for this - I'll add this in the next version which helps a lot because I am also going to be supporting SSL.

PR #96, opened 2017-07-25
Update material design icons to v1.9.32

minchik:


acockburn:
Thanks! Will be in next release

PR #99, opened 2017-07-28
Fixed issue with offset parameter in tutorial

linville:
Fixed issue with offset parameter in run_at_sunrise / run_at_sunset examples. Made the example slightly more complex (sunrise without `offset`, sunrise with a negative `offset`). Version 1.3.0 changed the offset parameter to a named parameter.


acockburn:
Thanks! I'll merge this in the next release.

PR #116, opened 2017-08-19
Add example apps for bysykkel, yr, ical, ruter, eventCache and eventMonitor

torkildr:
These apps are all a bit more "Hadashboard" centric than the other
examples. These apps are ported, or inspired by "jobs" used in the old
ruby version of Hadashboard.

They do not interact with Home Assistant at all, instead they fetch data
from external sources and publish them as Appdaemon events. These events
are are then consumed in dashboard widgets. Because we fetch and
tranform data in the apps, the widgets themselves will be pretty simple,
and simply format data, as it would any other Home Assistant event.

Apps:
 - bysykkel: Status of Oslo City Bikes
 - ical: Events from iCal feed. Perferct for Google Calendar .ics-feed
 - ruter: Oslo public transport real time data
 - yr: Detailed weather reports from Yr (more fine grained forecast than
   Home Assistant currenly have)
 - eventCache: Caching of events, to make dashboards behave better
   before data is loaded
 - eventMonitor: Debugging of events

### DIFF
--- a/conf/example_apps/bysykkel.py
+++ b/conf/example_apps/bysykkel.py
@@ -1,0 +1,45 @@
+import appdaemon.appapi as appapi
+import requests
+import json
+from datetime import datetime
+
+"""
+
+Get availability for Oslo City Bikes
+
+Arguments:
+ - event: Entity name when publishing event
+ - interval: Update interval, in minutes
+
+"""
+class Bysykkel(appapi.AppDaemon):
+    def initialize(self):
+        self.apiUrl = "http://reisapi.ruter.no"
+        self.entity = self.args['event']
+
+        now = datetime.now()
+        interval = int(self.args['interval'])
+
+        self.run_every(self.updateState, now, interval * 60)
+
+    def fetch(self, path):
+        res = requests.get(self.apiUrl + path)
+        return json.loads(res.text)
+
+    def updateState(self, kwargs=None):
+        status = self.getStatus()
+        self.set_app_state(self.entity, {
+            'state': "",
+            'attributes': status
+        })
+
+    def getStatus(self):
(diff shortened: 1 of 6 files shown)


homeassistant:
Hi @torkildr, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;appdaemon#116).

Thanks!

acockburn:
Hi there - sorry I missed this for some reason.

I Agree with @kellerza's suggested changes for the YR piece, and thanks for all the other examples. When you make the changes I will get these into the next version.

torkildr:
As far as I can understand from the terms and conditions for the usage of Yr, the usage of the data should be well withing their requirements, as long as the update interval is at least 10 minutes. This app does not actually have any configuration about interval, as that is in the appdaemon.yaml, but I can add a minimum update interval of 10 minutes.

The requirement of attribution is kinda tricky, seeing as this doesn't actually dipslay any information, and there's some interpretation about consuming data and "audience", but I'll also add a disclaimer text on the event object.

Do you have a link to the yr sensor discussion? I'm happy to add user agent information, but it should maybe be made clear that the calling clients are user owned installations

torkildr:
I've modified the pull request with the changes as stipulated above.

I've also created custom widgets corresponding to the apps in this request, which I will publish at later time, when widget API has matured a bit. I have added the attribution from yr dislaimer here, so as far as I can see, it should now be fully compliant.

kellerza:
Refer to https://github.com/home-assistant/home-assistant/issues/5154

acockburn:
Thanks!

PR #139, opened 2017-09-16
8 new widgets

ReneTode:


mmmmmtasty:
Would it be an idea to name the new_input_slider something like input_slider_smooth or input_slider_analog rather than using the word 'new'?

acockburn:
Yes, I'll be taking a look at that when I merge. I may well make the new input slider the default and make the existing one have a different name for folks who still prefer it.

ReneTode:
dont forget that renaming widgets are breaking changes.
i dont know if i would appreciate it when i have some dashboards made and suddenly after an update all my dashboards would look different.
its probably not the most used widget, so it still is possible, but i would be carefull with that.
from now on ill show you a pic and discuss naming with you before i create it ;)

PR #144, opened 2017-09-21
Added sticky flag to navigate component

larsenglund:
Added a stycky flag that will make the dashboard always return to the
dashboard in the args after a period of inactivity. Clicking the
dashboard resets (not cancels) the timeout. This is useful if you want a
main dashboard that the user always returns to after not using the
dashboard for a while.

Setting the sticky flag to 0 or omitting it will cause the default
timeout behaviour: to cancel the timeout when the user clicks something.

Example usage

```
args:
  timeout: 10
  sticky: 1
  return: main
```

### DIFF
--- a/appdaemon/assets/javascript/dashboard.js
+++ b/appdaemon/assets/javascript/dashboard.js
@@ -36,7 +36,16 @@ function ha_status(stream, dash, widgets)
                     {
                         ret = location.pathname
                     }
-                    timeout_params += "timeout=" + timeout + "&return=" + ret;
+                    if ("sticky")
+                    {
+                        sticky = data.data.sticky;
+                    }
+                    else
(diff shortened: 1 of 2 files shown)


homeassistant:
Hi @larsenglund, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;appdaemon#144).

Thanks!

larsenglund:
The name "sticky" and "1" and "0" might not be the best choice of name and value for this but it was what came out right on top of my head, feel free to change it to something better :)

acockburn:
Thanks for this - I am not ignoring it :) I will merge this soon after I have finished some refactoring.

PR #151, opened 2017-09-29
Fix style ovverrides related to basedisplay and sensor

mmmmmtasty:
This is an attempt to make the sensor widget obey the styles defined in it. It also adds the sub_state_map parameter which was missing previously. 

BUT

This actually doesn't quite work in the expected way - no matter what I do I cannot get "value_style" to apply to the "value" element. It always takes the value of "text_style" even though that is not referenced anywhere I can find.

For example you can see that with this config:

```
weather_hourly_summary:
  widget_type: sensor
  entity: sensor.dark_sky_hourly_summary
  value_style: "font-size: 60%; color: green"
  text_style: "font-size: 60%; color: red"
  state_text_style: "font-size: 60%; color: blue"
  unit_style: "font-size: 60%; color: purple"
```

The element still references the text_style not the value_style:

`<h2 class="value" data-bind="html: value, attr:{ style: value_style}" style="font-size: 60%;color: red;">Mostly cloudy until tomorrow afternoon.</h2>`

Where does this reference come from? I can also see that the fix applied to label.yaml in https://github.com/home-assistant/appdaemon/commit/9b2cc0613e3a53ce911dbeafe0c6722c5dc2fe86#diff-2c4b88e6573d551695bf8d006f064cf5 adds in text_style while also referencing basedisplay - again - how does that actually work and why doesn't that rely on value_style?


homeassistant:
Hi @mmmmmtasty, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;appdaemon#151).

Thanks!

mmmmmtasty:
Ignore my concerns, this change works as intended and is ready to merge, I tested on a new installation of appdaemon to confirm, I must have some caching issues on my old installation.

mmmmmtasty:
This was a poorly thought out breaking change, the text_style is intended to be applied to text (no duh!) while the value style is applied to numbers. The text sensor is obsolete, but either way I'm closing this.

PR #166, opened 2017-10-22
Fix syntax

pschmitt:
There was an extra `}` and the indentation was wrong.

By the way this file mixes tabs and spaces and lines end with `^M`. I didn't remove any of them, but wanted to ask beforehand.


acockburn:
Thanks for this I'll merge as is - if yo uwant to fix the rest of the stuff you mentioned that would be helpful :)

PR #201, opened 2018-01-08
Update sensor widget with container style

ReneTode:
create container style in sensor widget


ReneTode:
I think i was able to put the 7 PRs together to 1 PR.

PR #222, opened 2018-01-20
Fix cert_path key in call_service

denislooby:
I was having trouble getting Appdaemon and my self signed hass cert to work.
It could connect and read states from hass but any service call threw invalid SSL cert errors.
A bit of digging turned up that the call_service function was missing the underscore in cert_path.

Can also be worked around by adding a certpath key to the appdaemon.yaml with the same value as cert_path.

### DIFF
--- a/appdaemon/plugins/hass/hassapi.py
+++ b/appdaemon/plugins/hass/hassapi.py
@@ -413,8 +413,8 @@ class Hass(appapi.AppDaemon):
         )
 
         config = self.AD.get_plugin(self._get_namespace(**kwargs)).config
-        if "certpath" in config:
-            certpath = config["certpath"]
+        if "cert_path" in config:
+            certpath = config["cert_path"]
         else:
             certpath = None
 


homeassistant:
Hi @denislooby, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;appdaemon#222).

Thanks!

acockburn:
Thanks for fixing this - however, in beta 2 that I just released, I have deprecated cert_path. This is because I changed the HTTP client from requests to aiohttp which has better behavior in this area.  All you need to do is set `cert_verify` to `False` under the HASS plugin parameters and it will work fine with self signed certificates. This worked well for me when I tested against a self signed cert - let me know if this doesn't work for you.

denislooby:
@acockburn I didn't get a chance to update until now. All looking pretty good with my certs and the latest docker container.

Issue #246, opened 2018-03-01
Migrate from deprecated "async"


peterhoeg:
According to upstream:

>  DEPRECATION NOTE: Please do not use this library ! It’s not working correctly in python 3, and besides that might be called a failed experiment.

I don't know enough python to advise on a suitable substitute.

acockburn:
I'm not actually using it anywhere - where do you see it?

dotlambda:
It's in setup.py:
https://github.com/home-assistant/appdaemon/blob/415ae032e47a0eb5d4e1724a00d56b75207faccb/setup.py#L17

acockburn:
Thanks @dotlamba - I merged your PR and removed it.

@peterhoeg I'm not sure why that requirement was even there, I never used it to my knowledge. Thanks.

PR #252, opened 2018-03-04
Remove unused requirements

dotlambda:
Async seems not to be used anymore, see #246.
Configparser is part of the Python 3 standard library.


acockburn:
Ahh, I see what you mean now - not even sure how async got in there because I never used it - thanks.

PR #265, opened 2018-03-23
Fixed incorrect service call for input_text.set_value

snjoetw:


acockburn:
Thanks

snjoetw:
I just realized that both input_number and input_text are still present. For some reason I thought input_number is replaced by the latter. 

So the original behaviour is actually correct according to the doc. 

@acockburn fb22 What do you think we replace set_value to set_number_value and set_text_value?

fb22:
Fine by me.

PR #277, opened 2018-04-03
Update Apiai to DialogFlow and Support V2 API

engrbm87:
Google's API.AI has been renamed to DialogFlow and they have a new API V2 that has a different request and response format. So I have modified my original code contribution for this feature to reflect these changes.
- In the `get_dialogflow_intent()` and `get_dialogflow_slot_value()` funcs I changed the function to get the intent name and parameter values based on the used API version automatically. But for the response format function i had to manually specify the API version used in order to select the correct response format.
- I am not an expert in coding, so if you can help in passing the detected API version from the previous functions to the format response function that would be great.


engrbm87:
i am closing this pull request since the files have been updated since i created this PR. i will create a new PR based on the updated files.

PR #286, opened 2018-04-13
"its" vs. "it's"

smurfix:
Hi,
I do plan to submit more constructive patches in the future, but here's one for starters. ;-)

"its" is a possessive.
"it's" is shorthand for "it is".


mmmmmtasty:
This might be the most pedantic commit I've ever seen...I love it

Issue #305, opened 2018-05-05
"mode" widget broken for states like "on" and "off"


rschaeuble:
I have an input_select entity with (among others) the options "on" and "off".

I wanted to display this entity in a dashboard using the following widget definition:
```
some_state:
  widget_type: mode
  entity: input_select.entity_name
  mode: "on"
```

When loading the dashboard, the following error appears in the Javascript console: `Uncaught ReferenceError: True is not defined
    at HTMLDocument.<anonymous> (dashboard_name.js:33)`.

Clicking on the link in the stack trace reveals the following code:
```
... = new baseswitch("default-some-state", "", "default", {'widget_type': 'baseswitch', 'entity': 'input_select.entity_name', 'state_active': True, ...
```

The issue here is the `True` in `'state_active': True`. The value that should appear here instead of `True` is the `on` from my widget definition. The reason it doesn't is the following line of `mode.yaml`:
```
state_active: {{mode}}
```

Given my widget definition, this evaluates into `state_active: on`. As the `on` is not quoted, the YAML parser interpretes that as `state_active: True`.

The fix is rather easy: quote the right side (`state_active: "{{ mode }}"`).

If I'm not missing something, it should be safe (and useful) to go through all derived widgets and put all right sides with Jinja expressions in quotes, except for those where the right side is supposed to be a boolean.

I'll gladly fix this if I get confirmation that this is the way to go.

rschaeuble:
@acockburn: I have submitted a PR (#311). Please have a look and give me feedback whether you agree with that way of fixing the issue.
Before merging #311, please check #310 first, as #311 builds on top of #310.

acockburn:
Closing this as I have the PRs

PR #310, opened 2018-05-15
Yaml loading fix/cleanup

rschaeuble:
While working on #305 I found an error in logging YAML parsing errors.

I have fixed these, and also removed some small code duplication around parsing YAML files and handling errors.

PR #311, opened 2018-05-19
Yaml quotes

rschaeuble:
Fix for issue #305.

Until now, the yaml files were read line by line, replacing variables ("{{foo}}") on the fly.
If those variables contained quotes, that would easily result in an unparseable yaml file.

With this change, the yaml files are first parsed as-is. Afterwards, the resulting yaml document is traversed, replacing variables in strings. This avoids the issues with quotes completely.


rschaeuble:
This pull request includes pull request #310, so it might be best to review that one first.

rschaeuble:
@acockburn: did you have a chance already to look at this pull request (and at #310)?

PR #323, opened 2018-06-05
Add second positional argument `*args` to examples

TheBestMoshe:
When running the example code, I kept getting an error `TypeError: night() takes 1 positional argument but 2 were given`. Adding another positional argument fixed the problem. Therefore I added `*args` to all the example code on this page.

### DIFF
--- a/docs/TUTORIAL.rst
+++ b/docs/TUTORIAL.rst
@@ -139,10 +139,10 @@ different scenes in a different version of the App.
         self.run_at_sunrise(self.sunrise_cb)
         self.run_at_sunset(self.before_sunset_cb, offset=-900)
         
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @TheBestMoshe, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;appdaemon#323).

Thanks!

ReneTode:
there is no need to add args to the callbacks
there must be another reason why you get that error.

acockburn:
As Rene said - this shouldn't be necessary - hit us up on the forum if this is still an issue for you.

PR #327, opened 2018-06-14
Skin suitable for small screens (480x320 for example)

mad-ady:
It overrides some absolute positioning in some components in order for the layout to work. Here's a screenshot of it in action: https://imgur.com/VwRcMt8

Here's an article about it as well (on 2.x, though this version works on 3.x as well): https://magazine.odroid.com/article/home-assistant-designing-fancy-dashboard/


mad-ady:
I added a separate commit for the weather widget to display min/max temperature. It now looks like this: https://imgur.com/uE2SmW5.png

Questions:
1. how do I split this in 2 distinct pull requests? I can't seem to figure it out
2. Is there a way inside the baseweather widget to decide (via configuration) what gets displayed? e.g. maybe I don't want to display the row with apparent temperature and humidity.
3. What would be the best way to have different colors for low/high temperature available via configuration options?

Thanks.

ReneTode:
the skin is just a custom_skin and if i am correct then Andrew has said that he wouldnt put in more skins.
everyone can use it as it is.

the min/max in the weather was broken by an HA change, but with the options from the widget everyone can change it. still its usefull to update that.

1) sorry cant help you there)
2) not at the moment. but its on the list to implement such a thing
3) i guess by changing the css from the (custom) widget, because it isnt split up right now, if i am correct.

mad-ady:
Ok, thanks. Is there a skin repository where people can contribute skins? I think there should be, otherwise it would get lost on the internet and more people would end up reinventing the same skin.

Even if the temperature widget will not be merged, at least this change should be, in order to fix recent changes in HA with regard to dark sky: https://github.com/home-assistant/appdaemon/pull/327/commits/153d6ed0441f1e4058510c9edb7577b424c882ed#diff-4e2ba1407ae2c74d8dfdb1e5af39da32

ReneTode:
no there is no repository for skins.
there is no collection for skins setup at all at the moment.
if you dont want it to get lost then the best way is to share in on the forum, with a title like
HADashboard skin: some description.

im sure the fix will be merged in the next release.

PR #351, opened 2018-07-27
Update hassapi.py

Odianosen25:
Added namespace to the `self.get_state()` function within `self.set_app_state()`. Due to it not being there, it wasn't possible to properly edit an AppD entity in a different `namespace` outside that of Hass using `self.set_app_state()`, since it couldn't get the entity due to the namespace missing in the `self.get_state`.

Also added the "replace" flag in `self.set_app_state()`. As it was sometimes necessary (at least for me) to have all the data in the entity changed (like if wanting to delete some keys within attributes), rather than being updated. So if the "replace" flag is given, and is `True`, rather than update the attributes, it will replace the entire attributes dictionary with what was given.

By doing the above, in the app, one can read all attributes, modify as needed and decide to replace everything like removing keys.

Regards

### DIFF
--- a/appdaemon/plugins/hass/hassapi.py
+++ b/appdaemon/plugins/hass/hassapi.py
@@ -175,8 +175,6 @@ class Hass(appapi.AppDaemon):
         return new_state
 
     def entity_exists(self, entity_id, **kwargs):
-        if "namespace" in kwargs:
-            del kwargs["namespace"]
         namespace = self._get_namespace(**kwargs)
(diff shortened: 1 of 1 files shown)


acockburn:
I am happy to merge this but can you please add updates to the docs.

Odianosen25:
Hello @acockburn, Elmm please how can I do that? As I don't know which document I am to update and where it is. Can you kindly point me to the right direction?

Also has the custom plugin folder been resolved, so I can also add it to the docs?

Regards

PR #371, opened 2018-09-26
docs/docker_tutorial: fix typos

simonvanderveldt:
Noticed a couple of typos whilst reading the Docker docs, figured it was easy enough to fix them :)

The way the markdown is written is kinda "heavy" btw, using `#` or `##` for headings is a lot quicker than having a row below the heading with `---` or `===`. Also all inline monospace text has double backticks where a single one is fine. Don't know why it was done this way, if i had to guess it's probably a conversion from rst? Would be nice to clean it up, I can create a PR if that's OK?

### DIFF
--- a/docs/DOCKER_TUTORIAL.rst
+++ b/docs/DOCKER_TUTORIAL.rst
@@ -8,7 +8,7 @@ About Docker
 
 Docker is a popular application container technology. Application
 containers allow an application to be built in a known-good state and
-run totally independant of other applications. This makes it easier to
+run totally independent of other applications. This makes it easier to
 install complex software and removes concerns about application
 dependency conflicts. Containers are powerful, however they require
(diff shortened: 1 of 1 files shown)


acockburn:
Thanks, I'll take all the help I can get with the docs. Yes it was a conversion from RST so if you can simplify it that would be gratefully received! I'll merge when you have signed the CLA.

simonvanderveldt:
> Thanks, I'll take all the help I can get with the docs. Yes it was a conversion from RST so if you can simplify it that would be gratefully received!

OK, I'll make follow-up PR with some changes.

> I'll merge when you have signed the CLA.

done! :)

acockburn:
Thanks !!

simonvanderveldt:
@acockburn any chance you could merge this? Then I can create the follow-up PR :)

PR #379, opened 2018-10-08
changed some small errors in the climate widget

ReneTode:
units is unit in the code
level2_style is also in the code

### DIFF
--- a/docs/DASHBOARD_CREATION.rst
+++ b/docs/DASHBOARD_CREATION.rst
@@ -1586,7 +1586,7 @@ Optional Arguments:
 -  ``title2`` - a second line of title text
 -  ``step`` - the size of step in temperature when fading the slider up
    or down
--  ``units`` - the unit symbol to be displayed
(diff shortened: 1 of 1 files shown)


docs/DASHBOARD_CREATION.rst
```diff
@@ -1586,7 +1586,7 @@ Optional Arguments:
 -  ``title2`` - a second line of title text
 -  ``step`` - the size of step in temperature when fading the slider up
    or down
--  ``units`` - the unit symbol to be displayed
+-  ``unit`` - the unit symbol to be displayed
```
borpin: For consistency with `basedisplay` it should be `units`.

borpin:
There is also a `unit_style` available for modification.

docs/DASHBOARD_CREATION.rst
acockburn: Changed it to `units` in the widget so not merging this.

PR #386, opened 2018-10-15
option to connect to a local UNIX socket

benleb:
## Description:
Counterpart to the home-assistant/home-assistant#17489 which provides an UNIX socket.
This adds the ability to create/connect to a local UNIX socket. Connecting via a UNIX socket creates less overhead than TCP and also gives a little speed boost (maybe barely noticeable in many setups). Originally the idea was to circumvent the docker networking layers by using a UNIX socket which can be mounted into containers as a docker volume. I am using this code since 3-4 days to run my appdaemon apps and everything ran smoothly in this time. I also created the needed PR for appdaemon, see (will be updated in a second).

**Pull request in [home-assistant](https://github.com/home-assistant/home-assistant): home-assistant/home-assistant#17489
**Pull request in [appdaemon](https://github.com/home-assistant/appdaemon): https://github.com/home-assistant/appdaemon/pull/386


acockburn:
Thanks for this - can you provide an update for the docs too please?

benleb:
done :)

After some more testing, this maybe could also be mentioned in the docker guide 👍

acockburn:
Given the discussion on the HASS PR, is this still valid?

benleb:
yes :) I am working on a unix socket component which hopefully does not require any changes at the Hass HTTP component. Should be ready soon!

acockburn:
If that is the case, you can do the same with AD - you could create a custom plugin that uses sockets instead of the existing plugin.

benleb:
I will stay with patching hassplugin.py after every update. A whole plugin just to modify the connection part of the Hass plugin instead of changing/patching ~5 Lines in the file seems a little crazy/overkill. (would this be possible after all without touching the hass plugin?)

acockburn:
You could write a custom plugin - basically copy the hass plugin files to a dir called "custom_plugins" in your conf dir, - copy the whole "hass" directory of the plugin there and it will be used in preference to the built in hass plugin. That way at least you can upgrade AD without overwriting your changes, and could patch the plugin as necessary to keep up. 

Only thing you would need to change would be the import statement in your apps to:

```
import hass.hassapi
```

I'll be making a change soon to the plugin handling so that will be the same for apps using the built in plugin.

Use of custom plugins will be documented soonish.

PR #403, opened 2018-10-29
Adding support for json style args on widgets

clyra:
Trying again. I hope that this time I managed to did it right :-).

### DIFF
--- a/appdaemon/rundash.py
+++ b/appdaemon/rundash.py
@@ -346,6 +346,10 @@ class RunDash:
                         x = m.group(1)
                         y = m.group(2)
                         args["xy_color"] = [x, y]
+                elif key == "json_args":
+                      json_args = json.loads(data[key])
+                      for k in json_args.keys():
+                         args[k] = json_args[k]
                 else:
                     args[key] = data[key]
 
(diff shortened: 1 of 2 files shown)


acockburn:
I like this - but could you please add a section to the docs explaining it? If you can get that done soon I will get this into 3.0.2 which is ready for release in the next couple of days. Thanks!

clyra:
Yes, I think I can do that. But will need a little help and probably some gramar/spell checking. What you think about including around line 720 of WIDGETDEV.rst this:

Usually HADasboard understands ``args`` values as a single string. If you need to use a service which expects to receive a list or a dictionary then you may use the special key ``json_args`` and set its value to a stringified json. For example, suppose you want to pass to the service a list called ``colors``, then you could change the above code and include another check:

.. code:: javascript
            if ("my_json" in self.parameters)
            {
               args["json_args"] =  JSON.stringify(self.parameters.my_json);
            }

The corresponding widget configuration may include something like this:

.. code:: yaml

    my_json:
       colors:
         - red
         - blue
         - green

acockburn:
That looks good to me :)

clyra:
Done. Just committed the changes to doc.

acockburn:
Thanks!

PR #417, opened 2018-11-08
Gimme back the emoji!

benleb:
Just kidding... This adds a switch to disable the encoding of every log message to ascii. As someone (a german for example) might want to log "Umlaute" (äöü). Also useful if you want to place some christmas emoji in your log messages... 🎄🎅😄

### DIFF
--- a/appdaemon/utils.py
+++ b/appdaemon/utils.py
@@ -150,7 +150,7 @@ def single_or_list(field):
     else:
         return [field]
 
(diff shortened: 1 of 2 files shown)


acockburn:
Sorry to steal them - I obviously didn't realize what I was doing ;)

thundergreen:
Do i have to use dev version of appdaemon? Using hassio. sorry for dumb question

PR #433, opened 2018-11-22
add note on how to get an access token during initial setup with docker

rwa:
Setup note for people who aren't following all the details of HA development.


acockburn:
Thanks!

PR #464, opened 2019-01-03
Added optional default=None parameter to get_state().

efficiosoft:
Its value is returned for missing entities/attributes. This is handy when you e.g. expect a ``float`` to be returned and need to account for missing entities.

Without this, you have to do:

    get_state("climate.living", attribute="temperature") or 20

to be sure you get a ``float``, but you can't easily distinguish between ``0``, ``False`` and ``None``.

What do you think about it?

### DIFF
--- a/appdaemon/adapi.py
+++ b/appdaemon/adapi.py
@@ -319,12 +319,12 @@ class ADAPI:
         self.logger.debug("Calling info_listen_state for %s",self.name)
         return utils.run_coroutine_threadsafe(self, self.AD.state.info_state_callback(handle, self.name))
 
-    def get_state(self, entity_id=None, attribute=None, **kwargs):
+    def get_state(self, entity_id=None, attribute=None, default=None, **kwargs):
         namespace = self._get_namespace(**kwargs)
         if "namespace" in kwargs:
             del kwargs["namespace"]
 
-        return utils.run_coroutine_threadsafe(self, self.AD.state.get_state(self.name, namespace, entity_id, attribute, **kwargs))
+        return utils.run_coroutine_threadsafe(self, self.AD.state.get_state(self.name, namespace, entity_id, attribute, default, **kwargs))
 
     def set_state(self, entity_id, **kwargs):
         self.logger.debug("set state: %s, %s", entity_id, kwargs)
--- a/appdaemon/state.py
+++ b/appdaemon/state.py
@@ -276,13 +276,15 @@ class State:
 
         await self.AD.events.process_event(namespace, data)
 
-    async def get_state(self, name, namespace, entity_id=None, attribute=None):
-        self.logger.debug("get_state: %s.%s", entity_id, attribute)
+    async def get_state(
+            self, name, namespace, entity_id=None, attribute=None, default=None
+    ):
+        self.logger.debug("get_state: %s.%s %s", entity_id, attribute, default)
         device = None
         entity = None
         if entity_id is not None and "." in entity_id:
             if not await self.entity_exists(namespace, entity_id):
-                return None
+                return default
         if entity_id is not None:
             if "." not in entity_id:
                 if attribute is not None:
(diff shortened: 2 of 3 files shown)


efficiosoft:
@ReneTode Sorry, but I have to disaggree in multiple points.

Why should be taking a fallback value for entities that maybe haven't yet loaded when Home Assistant has just started be something not wanted?

There are similar usecases all around Pythons stdlib that follow the same pattern and are used very widely, e.g. ``min`` or ``max``, which take a ``default`` argument with a value to return when zero values are given to calculate the min/max from. Or ``dict.get``... the examples are numerous.

Next, this

    get_state("climate.living", attribute="temperature") or 20

is not a perfect solution, because the ``or`` would catch not only when ``get_state()`` returns ``None``, but also for ``0``, ``False``, ``None``, empty strings, empty lists and any other object whichs ``__bool__()`` returns ``False``, as I already mentioned.

But nevertheless, that particular issue could be solved with a custom helper that rewrites ``None`` to a custom value, it's just not ellegant.

What is a far bigger problem is that you are not able to distinguish between an attributes value that is set to ``None`` (yes, that's possible, states are just JSON) and a missing attribute. The only really right way to deal with this would be raising an exception for non-existing entities/attributes, but that would break backwards-compatibility, hence I didn't suggest it at all. But being able to specify a fallback value is a acceptable tradeoff IMO.

Best regards
Robert

ReneTode:
> Why should be taking a fallback value for entities that maybe haven't yet loaded when Home Assistant has just started be something not wanted?
there is no way to determine if it isnt loaded yet or if it does exist.
something that doesnt exist should never reurn a state in my opinion. that would lead to behaviour you cant see and cant control.
you wouldnt be able to tell if the entity is not existing and returning a state or if it is actually an existing entity that returns a valid state. get_state would be an independable function.

> is not a perfect solution, because the or would catch not only when get_state() returns None, but also for 0, False, None, empty strings, empty lists and any other object which's __bool__() returns False, as I already mentioned.
then add the existing failsaves that are reliable, to make sure you get the right value.

```
if self.entity_exists("climate.living"):
  a = get_state("climate.living", attribute="temperature")
else:
  a=20
```
then you are sure you get the value from an existing entity or 20

efficiosoft:
> something that doesnt exist should never reurn a state in my opinion. that would lead to behaviour you cant see and cant control.

Well, then you should disaggree with the current implementation as well, because it returns something for missing attributes: ``None`, which can be a valid attribute's value`. You would in fact gain more control when you're able to specify a fallback, because you could specify a value you know the attribute will never have and then determine if it exists or not. As I said, the proper way of handling this would be exceptions, but I didn't consider them here for said reasons.

And yes, I know you can retrieve the whole entity state and check whether it contains the attribute in question, but that's just not straightforward.

ReneTode:
> Well, then you should disaggree with the current implementation as well, because it returns something for missing attributes: None

good point. i guess i never deal with that because i never have attributes that have the state None.
and when i create attributes or states i avoid None also (and use "" instead)

i guess it should not return None also and raise an error.

> And yes, I know you can retrieve the whole entity state and check whether it contains the attribute in question, but that's just not straightforward.

if i program for someone else i always put in checks if they have given valid arguments, so if the attribute or entity exists, if i program for myself i make sure that i only use elements that exist.

i cant think of any situation where i want to get a value in return, when i made a typo in the attribute name, and then my app goes on with that value. if i get None i know something is wrong and if i know i can expect None, i would for sure add double checking.

efficiosoft:
Well, there are quite multiple thinkable scenarios in which I would like to determine if an attribute exists or is ``None``, or  I just want to use a fallback value in case an entity doesn't exist.

As it stands right now, I see two possible solutions to this:

1. Raise an exception when the requested attribute/entity doesn't exist - would be backwards-incompatible.
2. Just adding the default parameter. This is backwards-compatible, makes differenciating between missing attributes/entities and those just ``None`` possible and provides a way to have a fallback value.

It's not that it couldn't be done manually as well, it's just a huge simplification, equivalent to

    try:
        value = dict[key]
    except KeyError:
        value = 25

vs just

    value = dict.get(key, 25)

acockburn:
This looks OK to me - it doesn't break anything and adds to the usability of the calls so I'll merge it for the next release - thanks!

ReneTode:
> But let's wait what @acockburn sais about it. I think I've explained my standpoint now.

i guess you know now. he convinced me that only those people who would want to use it would run into trouble if they are not carefull ;)

efficiosoft:
@ReneTode  and @acockburn  Thanks for the discussion and accepting it. I'd normally open an issue to discuss before implementing such things, but I already had this laying around and it was a rather small change, so please be lenient with me :)

acockburn:
Closing this as commit is included in #468

efficiosoft:
Well, depends on whether #468 will be accepted. I based it on this branch because you approved it yesterday and it would have caused merge conflicts otherwise.

acockburn:
I'll be accepting it :)

efficiosoft:
Ok, thanks ;)

PR #468, opened 2019-01-05
    Added copy=True parameter to get_state()

efficiosoft:


efficiosoft:
Ah, this is based on the changes from #464.

PR #478, opened 2019-01-15
Ability to Access HA's History

Odianosen25:
Added new changes that allows for service calls to the HA history/database
This will also allow for external REST access to AD's REST API


acockburn:
Please add docs for this

PR #482, opened 2019-01-17
Made docker image more compact (merged layers, purged cache)

efficiosoft:
This doesn't change any functionality, but combines build steps for the docker image, resulting in fewer image layers and thus smaller size and slightly improved performance.

Can also be backported to 3.x, if desired.

PR #486, opened 2019-01-19
Change Docker base image to alpine linux

benleb:
Hey, this PR should start a discussion if a full blown debian as base image for appdaemon is really needed?! 
The Dockerfile in this PR uses alpine linux as base image which reduces the size dramatically (~380MB vs. 52MB) and probably also improves the performance (to build and run) slightly (just my personal feeling, no tests or numbers about this).

I added _curl_ to the image as it is often used by Docker healthchecks and not available in a plain alpine linux. Via this mechanism and for example Docker Build-Args, custom alpine packages (and also pip ones) could easily be added by the user if needed.
The changes to the _.dockerignore_ mentioned by @efficiosoft in #482 are also integrated here, thanks!
Python 3.7 is used here (for personal reasons) but it was also built and tested with Python 3.6 without any problems.

I personally use this since few weeks and have not seen any negative effects right now. But as this highly depends on the apps run, more testers are needed ;)

Images for quick testing are available on [dockerhub benleb/appdaemon-alpine](https://cloud.docker.com/repository/docker/benleb/appdaemon-alpine) for the current _dev_ and _3.0.2_ or can be built with your favorite _docker build_ command.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,9 @@
 .git
+.gitignore
+
 Dockerfile
+.dockerignore
+
+*.md
+*.rst
+docs
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,4 @@
-FROM python:3.6
-
-VOLUME /conf
-VOLUME /certs
-EXPOSE 5050
+FROM python:3.7-alpine
 
 # Environment vars we can configure against
 # But these are optional, so we won't define them now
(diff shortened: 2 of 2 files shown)


efficiosoft:
Hi,

I deem switching to Alpine a good idea, as well as including curl.

However, I'm not sure about the proper strategy to install user-specific Python packages. While this can be done with build args, and I clearly see the advantages of doing it this way, it is probably not feasible to recommend an average docker user building his own image when he just wants to use third-party apps from PyPi.., because lots of users don't have proper management tools in place and will probably just set and forget it, missing any future updates to the base image.

There exists a mechanism for installing packages from PyPi via requirements.txt files in the apps directory, which is quite easy to use. However, the current issue with this method is that the installed packages don't survive a container restart/recreation and container startup fails when there's no internet connectivity, which clearly is a No-Go. I'm already working on this and wanted to start a discussion about this myself, but maybe we can discuss this here as well, since it's all about the Docker deployment method.

So my idea was to establish a local store of wheel files inside the config directory (or probably better a separate volume), which are tried to be updated when there's internet connectivity, and if there's not, are just installed upon container startup. I've written a script for this purpose (https://github.com/efficiosoft/hass-apps/blob/wheelstore/wheelstore.py), which would need to be invoked with ``establish`` and then ``install`` from the docker start script. I think I prepared this already, but it needs some cleanup before I can make a PR.

What do you think? I hope you don't see this as thread hijacking, but as it's related to the proposed build args to install packages, I just put my two cents in.

efficiosoft:
You're right, let's keep this separate.

So basically, yes, I don't see anything standing against allowing installing additional packages via build args, although this is probably not suited for the average user. And those who are familiar with Docker will create their own child image anyway, so that they can take additional steps after installing the packages.

PR #518, opened 2019-02-12
Removed mqtt.none that was created by default

Odianosen25:
Removed an entity which was usually created by default in the namespace.


acockburn:
Thanks!

PR #524, opened 2019-02-23
code example type-o

bbrendon:
I've noticed these over the years in the docs. I think there are other similar issues but figured I'd start with this one.


acockburn:
Thanks - we'll get this into 4.0

Issue #525, opened 2019-02-25
Using the '__function__' flag in Logs


Odianosen25:
When trying to log with the function name being used like

```
def some_function(self, kwargs):
    self.log('__function__, __line__')
```
`__line__` returns the wrong line as `86`, so also `__function__` returns `log` instead of the function name `some_function`.

Regards

Odianosen25:
Fix here #526

PR #526, opened 2019-02-25
Fix for Log Function and Line issue

Odianosen25:
This fix is for the log issue #525 where by when `__function__` or `__line__` is used, it keeps giving `log` and `86` all the time


acockburn:
Thanks!

acockburn:
This change was included in your delete entity patch so closing.

PR #528, opened 2019-02-26
fix kwargs handling by using pop instead of get

benleb:
without popping 'ascii_encode', it gets hand over to the python logger which then complains about an unexpected keyword...

```
  ...
  File "/usr/local/lib/python3.7/logging/__init__.py", line 1449, in log
    self._log(level, msg, args, **kwargs)
TypeError: _log() got an unexpected keyword argument 'ascii_encode'```


acockburn:
Thanks!

PR #530, opened 2019-03-06
Added Shutdown payload

Odianosen25:
Updated document to add the new MQTT config feature for the ability to send a message when AD disconnects cleanly from the broker


acockburn:
Thanks!

PR #531, opened 2019-03-14
Fix run_once when comparing event with now

engrbm87:
The event datetime object is naive in the run_once function. It needs to be converted to aware_event before comparing it with now.


acockburn:
Thanks! I'll get this into the next release.

PR #534, opened 2019-03-21
Update to Save_namespace and Remove_entity

Odianosen25:
Added key worded arguments to `save_namespace`, so once the app level namespace is declared using `set_namespace`, when using `self.save_namespace`, no need to start declaring the namespace again. 

Also if another namespace is required, by simply following other AD conventions of declaring namespace `namespace = namespace` , its use becomes in line with other AD apis.


Odianosen25:
Also added the ability to remove an entity

PR #558, opened 2019-05-02
Different fixes and changes

Odianosen25:


Odianosen25:
oh but I followed the same process as before. 

So what should I do to clean it up?

Reards

acockburn:
What are you actually trying to do with this PR? As far as I can tell, it is just a collection of PRs you previously merged. Which out of all the commits are stuff you want to be in the PR?

PR #560, opened 2019-05-05
update docs for Raspberry Pi Docker instructions

tseho:
I updated the **Raspberry Pi Docker** instructions on `docs/INSTALL.rst` since they were outdated. 

The most important change is to use `FROM arm32v7/python:3.6-alpine` instead of `FROM arm32v7/python:3.6` because the image has been changed to alpine in #486.

### DIFF
--- a/docs/INSTALL.rst
+++ b/docs/INSTALL.rst
@@ -209,32 +209,33 @@ Raspberry Pi Docker
 -------------------
 
 Since the official Docker image isn't compatible with raspberry Pi, you will need to build your own docker image
-from the downloaded repository. The Dockerfile also needs a couple of changes:
+from the downloaded repository.
 
-1. Change the image line to use a Resin image:
+.. code:: bash
 
-``FROM arm32v7/python:3.6``
+    $ git clone https://github.com/home-assistant/appdaemon.git
+    $ cd appdaemon
 
-2. Change the ``RUN`` line to the following:
+You must change the image line to use a Resin image in the Dockerfile:
 
-``RUN pip3 install requests && pip3 install .``
+``FROM arm32v7/python:3.6-alpine``
 
 You can then build and run a docker image locally as follows:
 
 .. code:: bash
-    $ git clone https://github.com/home-assistant/appdaemon.git
-    $ cd appdaemon
+
     $ docker build -t appdaemon .
-    $ docker run -t -i --name=appdaemon -p 5050:5050 \
-      -e HA_URL="<Your HA URL>" \
-      -e HA_KEY="<your HA Key>" \
-      -e DASH_URL="<Your DASH URL>" \
-      -v <Your AppDaemon conf dir>:/conf \
+    $ docker run --name=appdaemon -d -p 5050:5050 \
+      --restart=always \
+      -e HA_URL="<Your HA_URL value>" \
+      -e TOKEN="<your TOKEN value>" \
+      -e DASH_URL="http://$HOSTNAME:5050" \
+      -v <your_conf_folder>:/conf \
       appdaemon:latest
 
 For more information on running AppDaemon under Docker, see the Docker Tutorial. The key difference is that
 you will be running a locally built instance of AppDaemon rather than one from Docker Hub, so for run commands,
-make usre yo uspecify "appdaemon:latest" as the image, as above, rather than "acockburn/appdaemon:latest" as the tutorial states.
+make sure to specify "appdaemon:latest" as the image, as above, rather than "acockburn/appdaemon:latest" as the tutorial states.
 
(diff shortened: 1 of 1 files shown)


Erovia:
Since Docker now supports multi-arch images, separate Dockerfile for ARM builds is no longer necessary.
`FROM python:3.7-alpine` is perfectly usable for non-x86_64 architectures.

It'd be best, if the project provided official builds for at least arm32v7, too.

tseho:
@Erovia Thanks for the info, I will look into it tomorrow and if you are right, I will update my PR with the appropriate documentation.

acockburn:
@tseho - thanks for this. Is your update still pending or shall I merge as is?

tseho:
@acockburn I haven't tried yet, the test should be quick, I will launch it tonight and will get back to you asap.

tseho:
@Erovia was right, there is no need to specify the `arm32v7/python:3.7-alpine`, `python:3.7-alpine` is working as it is on arm32v7.
I will update my PR accordingly.

tseho:
@acockburn I updated it.

Erovia:
@acockburn, do you think it would be possible the provide armhf builds of AppDaemon on your Docker Hub repo? It can be an automated build, with some extra work: https://github.com/docker/hub-feedback/issues/1261

acockburn:
thanks @tseho 

@Erovia - yes it would make sense to do this. I'll take a look when I have some spare time.

tseho:
@Odianosen25 I'm sorry to hear that. I want to be sure it's working for everyone.

Can you provide more informations so I can try to reproduce the error ?
Which device ? Can you copy the docker build logs and docker logs when trying to start it ?

Odianosen25:
@tseho, 

Well I am not sure where it messed up, as I am not a docker user, but had to check that it works.

So as I said it all built fine, but I think the issue is just my own system as my code needed `psutil` installed and it came up with 

```
 -g -fwrapv -O3 -Wall -DTHREAD_STACK_SIZE=0x100000 -fPIC -DPSUTIL_POSIX=1 -DPSUTIL_VERSION=562 -DPSUTIL_LINUX=1 -DPSUTIL_ETHTOOL_MISSING_TYPES=1 -I/usr/local/include/python3.7m -c psutil/_psutil_common.c -o build/temp.linux-armv7l-3.7/psutil/_psutil_common.o
2019-05-22T10:13:05.454637382Z     unable to execute 'gcc': No such file or directory
2019-05-22T10:13:05.454665924Z     error: command 'gcc' failed with exit status 1
2019-05-22T10:13:05.454693319Z     ----------------------------------------
2019-05-22T10:13:05.491134099Z ERROR: Command "/usr/local/bin/python -u -c 'import setuptools, tokenize;__file__='"'"'/tmp/pip-install-rlsmvt2b/psutil/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /tmp/pip-record-vjy_gd7w/install-record.txt --single-version-externally-managed --compile" failed with error code 1 in /tmp/pip-install-rlsmvt2b/psutil/
```
This I wouldn't imagine affects others, so based on that didn't think there was any need stopping merging it.

If I was fluent in docker, maybe I can. But unfortunately I am not. So if you can point me in the right direction, that will be great.

Thanks

tseho:
@Odianosen25,

Assuming the official docker image for appdaemon will support the multi-arch builds in the future, I will not recommend to change the original Dockerfile.

What I would do, if I were in your case, would be the following:

```
$ git clone https://github.com/home-assistant/appdaemon.git
$ cd appdaemon
$ docker build -t appdaemon .
```

Once the hack for building the official docker image on the arm32v7 arch is done,

Then, I would create a custom Dockerfile in another directory, eg: `odianosen/Dockerfile`:
```
# Here, we are reusing the arm32v7 build of the docker image previously created
FROM appdaemon:latest

RUN apk add python3-dev
```

Now, you can build/run the custom image like this :
```
$ docker build -t odianosen/appdaemon .
$ docker run --name=appdaemon -d -p 5050:5050 \
      --restart=always \
      -e HA_URL="<Your HA_URL value>" \
      -e TOKEN="<your TOKEN value>" \
      -e DASH_URL="http://$HOSTNAME:5050" \
      -v <your_conf_folder>:/conf \
      odianosen/appdaemon:latest
```

This is a specific use case, I don't think we should pollute the documentation with it :wink:

Odianosen25:
Thanks @tseho,

I agree with that, and that’s why I didn’t bother and merged. 

Thanks for the update and kind regards

PR #567, opened 2019-05-19
Added Sensors and Doc update

Odianosen25:
- Added three sensors for "total", "active" and "inactive" apps
- Fix for wrong total number of apps calculated
- Fix for reloading of nested app configuration 
- Added extra logging for total and active apps
- Ensured `admin` namespaces cannot be listened to for now
- Some doc update with some misspelling

### DIFF
--- a/appdaemon/app_management.py
+++ b/appdaemon/app_management.py
@@ -68,11 +68,11 @@ class AppManagement:
     async def remove_entity(self, name):
         await self.AD.state.remove_entity("admin", "app.{}".format(name))
 
-    async def entity_exists(self, namespace, entity_id):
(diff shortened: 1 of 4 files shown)


ReneTode:
hmm, odia didnt we have a discussion about this?

Odianosen25:
Yes but it had been added already.

In the latest PR I did, I removed most of it, especially the listing of the apps in the sensor for memory as we discussed.

So all it will do, is just have a sensor which is populated once and that’s all. No constant checking as I did before.

Now u mentioned it, just occurred to me I had nothing written abt it in the PR 😬. Glad Andrew hasn’t seen it 😏

PR #568, opened 2019-05-22
Added App Management API Calls

Odianosen25:
Added api calls to manage apps. These include
- Starting apps
- Stopping apps
- Restarting apps
- Registering App Services
 
Also added documentation for the above


Odianosen25:
@acockburn,

Changed to `managed_services` boss

PR #579, opened 2019-06-06
Modified `immediate` Flag in Listen_state

Odianosen25:
Expanded on the ability to allow for when listening to state changes, the callback function can be ran immediately to allow for the function to be ran at startup.

The following are possible:
- If `duration` is not specified, the callback is ran immediately. If specified, it will be ran after the specified `duration`
- If `new` is specified, it will only run the callback, if the present `state` is the `new` state

Regards

### DIFF
--- a/appdaemon/adapi.py
+++ b/appdaemon/adapi.py
@@ -514,6 +514,7 @@ class ADAPI:
 
     def cancel_timer(self, handle):
         name = self.name
+        self.logger.debug("Canceling timer with handle %s for %s", handle, self.name)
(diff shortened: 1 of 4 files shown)


ReneTode:
please check for typos

Odianosen25:
Done that :) thanks

ReneTode:
If ``new`` and ``duration``is not set should be
If ``new`` and ``duration`` are not set

Odianosen25:
Fixed thanks

PR #589, opened 2019-07-03
Fix incorrect name of event fire on disconnect in docs

rsokolowski:
Fix the name of the event that gets fired by AD when plugin gets disconnected - the actual name of the event is "plugin_stopped" as defined https://github.com/home-assistant/appdaemon/blob/fbdd4f1242076d849b6d764dd82e362ffb729641/appdaemon/plugin_management.py#L185


homeassistant:
Hi @rsokolowski, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;appdaemon#589).

Thanks!

PR #619, opened 2019-08-14
Add Support for rendering HASS templates

dlashua:
This provides self.render_template(template="some hass template here").

To be used like this:
```python


class Test(hass.Hass):

    def initialize(self):
        ret = self.render_template("{{ states('sensor.house_mean_temp') }}")
        self.log("FLOAT {}: {}".format(type(ret), ret))

        ret = self.render_template("{{ states('binary_sensor.office_occupied') }}")
        self.log("STRING {}: {}".format(type(ret), ret))

        ret = self.render_template("{{ is_state('binary_sensor.office_occupied', 'on') }}")
        self.log("BOOL {}: {}".format(type(ret), ret))
```

### DIFF
--- a/appdaemon/plugins/hass/hassapi.py
+++ b/appdaemon/plugins/hass/hassapi.py
@@ -1,4 +1,5 @@
 import requests
+from ast import literal_eval
 
 import appdaemon.adbase as adbase
 import appdaemon.adapi as adapi
@@ -279,3 +280,20 @@ class Hass(adbase.ADBase, adapi.ADAPI):
 
         result = self.call_service("database/history", **rargs)
         return result
+
+    @hass_check
+    def render_template(self, template, **kwargs):
+        namespace = self._get_namespace(**kwargs)
+
+        if "namespace" in kwargs:
+            del kwargs["namespace"]
+
+        rargs = kwargs
+        rargs["namespace"] = namespace
+        rargs["template"] = template
+
+        result = self.call_service("template/render", **rargs)
+        try:
+            return literal_eval(result)
+        except (SyntaxError, ValueError):
+            return result
--- a/appdaemon/plugins/hass/hassapi.py
+++ b/appdaemon/plugins/hass/hassapi.py
@@ -283,6 +283,28 @@ class Hass(adbase.ADBase, adapi.ADAPI):
 
     @hass_check
     def render_template(self, template, **kwargs):
+        """Renders a Home Assistant Template
+
+        Args:
+            template (str): The Home Assistant Template to be rendered.
+
+        Keyword Args:
+            None.
+
+        Returns:
+            The rendered template in a native Python type.
+
+        Examples:
+            >>> self.render_template("{{ states('sun.sun') }}")
+            Returns (str) above_horizon
+
+            >>> self.render_template("{{ is_state('sun.sun', 'above_horizon') }}")
+            Returns (bool) True
+
+            >>> self.render_template("{{ states('sensor.outside_temp') }}")
+            Returns (float) 97.2
+
+        """
         namespace = self._get_namespace(**kwargs)
 
         if "namespace" in kwargs:
--- a/appdaemon/plugins/hass/hassplugin.py
+++ b/appdaemon/plugins/hass/hassplugin.py
@@ -453,8 +453,12 @@ class HassPlugin(PluginBase):
 
             apiurl = "{}/api/history/period{}{}{}".format(config["ha_url"], timeStamp, filter_entity_id, eTime)
 
+        elif domain == "template":
+            apiurl = "{}/api/template".format(config["ha_url"])
+            
         else:
             apiurl = "{}/api/services/{}/{}".format(config["ha_url"], domain, service)
+
         try:
             if domain == "database":
                 r = await self.session.get(apiurl, headers=headers, verify_ssl=self.cert_verify)
@@ -462,7 +466,10 @@ class HassPlugin(PluginBase):
                 r = await self.session.post(apiurl, headers=headers, json=data, verify_ssl=self.cert_verify)
                 
             if r.status == 200 or r.status == 201:
-                result = await r.json()
+                if domain == "template":
+                    result = await r.text()
+                else:
+                    result = await r.json()
             else:
                 self.logger.warning("Error calling Home Assistant service %s/%s/%s", namespace, domain, service)
                 txt = await r.text()
@@ -570,6 +577,7 @@ class HassPlugin(PluginBase):
             r.raise_for_status()
             services = await r.json()
             services.append({"domain": "database","services": ["history"]}) #manually added HASS history service
+            services.append({"domain": "template","services": ["render"]})
 
             return services
         except:


ReneTode:
why at all would someone like to do that??
most people i know that use AD are using AD to get away from that kind of complex structures?
```
ret = self.render_template("{{ states('sensor.house_mean_temp') }}") # equals
ret = self.get_state("sensor.house_mean_temp")

ret = self.render_template("{{ states('binary_sensor.office_occupied') }}") # equals
ret = self.get_state("binary_sensor.office_occupied")

ret = self.render_template("{{ is_state('binary_sensor.office_occupied', 'on') }}")
ret = ( self.get_state("binary_sensor.office_occupied") == "on")
```
get_state does the trick better, and way easier.

dlashua:
The use case is something like this app configuration:

```
my_app:
  module: vacuum_handler
  class: VacuumHandler
  rooms:
    living:
      condition: "{{ is_state('binary_sensor.living_occupied', 'off') and not is_state('input_select.house_mode', 'sleep') }}"
    bedroom:
      condition: "{{ is_state('binary_sensor.bedroom_occupied', 'off') and not is_state('input_select.house_mode', 'sleep') }}"
```

AppDaemon has constraints, yes. But, constraints defined in the app YAML limit callbacks for the entire app. Additionally, even when managed in the module, additional code is needed to perform "and", "or", and numeric comparisons.

Faced with a need to allow an undetermined number of user-supplied (in the yaml, not in the module) constraints, my options were to roll my own everything, extend the constraint system, or leverage Home Assistant's templates. Home Assistant Templates got the job done quickly since Home Assistant already provides an API call for this. 

Ideally, there'd be a more "AppDaemon"-ish way to do this, which I plan to work on in the future. But this handled the need for now without disrupting anything else, and I thought someone else might have a need for this, too.

ReneTode:
> AppDaemon has constraints, yes. But, constraints defined in the app YAML limit callbacks for the entire app. Additionally, even when managed in the module, additional code is needed to perform "and", "or", and numeric comparisons.
constraints can be place on any listener or schedular (listen_state, listen_event, run_in, run_at, run_at_sunrise, etc) not only at app level.
and with custom constraints you can create your own custom condition in any way you like, and with every python function possible.

i myself have a custom constraint that i use like:
```
  hold_for:
    - sensor.rene: "In bed"
    - input_boolean.alarm_marjolijn: "off"
    - sensor.marjolijn: ["Thuis","uitkleden","wakker worden"]
    - time: {"start": "11:00:00", "end": "sensor.bedtijd_start"}
```
all within the possibilities from AD at this moment. and in the AD way.
https://appdaemon.readthedocs.io/en/latest/APPGUIDE.html#custom-constraints

this is a possible custom constraint that asks for the input i showed above:
```
    def custom_constraint(self, conditions, debug_logging, language):
        #########################################################################################
        # lets start by seeing if we need to translate:
        #########################################################################################
        #debug_logging = True
        if "no_hold" in conditions:
            if debug_logging:
                self.log("no hold for was set")
            return True
        self.language = self.app_config["actions_language_" + language]
        #self.log("hold conditions are: {}".format(conditions))
        for condition in conditions:
            for device,value in condition.items():
                #if debug_logging:
                #    self.log("hold for: {} value: {}".format(device,value))
                if device == self.language["weekday"]:
                    for _day in value:
                        if (datetime.datetime.today().weekday() + 1) == _day:
                            if debug_logging:
                                self.log("i will hold for: {} value: {}".format(device, _day))
                            return False
                elif device == self.language["time"]:
                    if "sensor" in value[self.language["start"]]:
                        starttime = self.get_state(value[self.language["start"]])
                    else:
                        starttime = value[self.language["start"]]
                    if "sensor" in value[self.language["end"]]:
                        endtime = self.get_state(value[self.language["end"]])
                    else:
                        endtime = value[self.language["end"]]
                    if self.now_is_between(starttime,endtime):
                        if debug_logging:
                            self.log("i will hold for: {}".format(device))
                        return False
                elif isinstance(value, list):
                    for eachvalue in value:
                        if str(self.get_state(device)) == str(eachvalue):
                            if debug_logging:
                                self.log("i will hold for: {} value: {}".format(device,eachvalue))
                            return False
                elif str(self.get_state(device)) == str(value):
                    if debug_logging:
                        self.log("i will hold for: {} value: {}".format(device,value))
                    return False
                if debug_logging:
                    self.log("didnt hold for: " + device)
        return True
```

so i think you found a way to do what already was possible, but that you didnt know about.

dlashua:
Right, constraints can be placed on any callback (scheduler call backs too). 

I appreciate you sharing the above code. Unfortunately, in my particular use case, it doesn't help. 

And, again, I know there are many, many ways to tackle this problem. This is just one way to do it. It doesn't have to be the way you would choose in order to be useful for someone else.

If you want to let people run Home Assistant templates from AppDaemon, then great, here's a PR. If not, that's okay too.

ReneTode:
like i said in the usecase you showed it would work.
indeed there are many ways to do things, that doesnt mean that they all need to be included in AD.

I am not the one deciding if it goes into AD or not, but Andrew.
but because its not really adding anything to AD, that cant be done without it, my personal view is that it shouldnt be added.

by the way after looking at the code its just a convienient function.
so all it does is changing this code (which everyone can use already):
```
self.call_service("template/render", template = "{{ states('sensor.house_mean_temp') }}")
```
into this
```
self.render_template("{{ states('sensor.house_mean_temp') }}")
```

when we start adding convienient functions for all available HA services, we would be going on for a long time.

dlashua:
> but because its not really adding anything to AD, that cant be done without it, my personal view is that it shouldnt be added.

Everything AppDaemon does I could do with a Home Assistant automation, or a custom_component. The very nature of AppDaemon (at least in part) is making it better/easier/faster. By your logic, all of AppDaemon shouldn't be added because one can already do all the things it does.

There are other use cases for rendering a Home Assistant Template in AppDaemon. On top of that, the use case I presented is still valid, because your solution was to share code you've written that isn't part of AppDaemon. But, it doesn't matter. If you don't see a use for this, then you don't. And if @acockburn doesn't either, that's, of course, fine. In true FOSS spirit, I write code and make changes because it suits me. And, in that same spirit, when I feel those changes are useful to others, I offer them back upstream. The project owner can decide if it suits the project. 

> so all it does is changing this code (which everyone can use already)

Not at all.

With my PR, This test code:
```
class TestRender(hass.Hass):

    def initialize(self):
        ret = self.call_service("template/render", template = "{{ states('sensor.house_mean_temp') }}")
        self.log(ret)
```

Produces this result:
```
2019-08-15 09:29:28.955513 INFO AppDaemon: Initializing app test_render using class TestRender from module test
2019-08-15 09:29:28.992794 WARNING AppDaemon: Unknown domain (default/template) in call_service
2019-08-15 09:29:28.996199 INFO test_render: None
```

ReneTode:
```
Everything AppDaemon does I could do with a Home Assistant automation, or a custom_component. The very nature of AppDaemon (at least in part) is making it better/easier/faster. By your logic, all of AppDaemon shouldn't be added because one can already do all the things it does.
```
that isnt true, because i can automate my house without HA at all, using only appdaemon.
so its HA that is slowly gettin obsolete for appdaemon users and not the other way around ;)


the code you wrote:
```
    def render_template(self, template, **kwargs):
        namespace = self._get_namespace(**kwargs)

        if "namespace" in kwargs:
            del kwargs["namespace"]

        rargs = kwargs
        rargs["namespace"] = namespace
        rargs["template"] = template

        result = self.call_service("template/render", **rargs)
        try:
            return literal_eval(result)
        except (SyntaxError, ValueError):
            return result
```
does nothing more then return self.call_service("template/render") with some args.
and thats a normal AD function.
but it shows that the call_service is incomplete because there are services in HA that are on a different place in the HA api.
and it shows that its possible to get response back from a service, which isnt the case now, i believe.

so my advice to @acockburn is to extend the call_service function so that it can be used for this also.

dlashua:
> that isnt true, because i can automate my house without HA at all, using only appdaemon.

And you can automate your house without AppDaemon at all, using only HA. We both agree that AppDaemon is great, that's why we use it. 


> so my advice to @acockburn is to extend the call_service function so that it can be used for this also.

I think you should look over the commits on this PR again. 

I committed changes to TWO files. 

1) hassapi.py which provides the convenience method, self.render_template(). 

2) hassplugin.py which extends the call_service function so it can be used for this, the very thing you're advising @acockburn to do.

I didn't commit a convenience function that doesn't also include the plugin code to make it work. That would accomplish nothing.

ReneTode:
i saw it, thats why i advised to change the call_service.
there might be even more services on other places in the HA api though.

i personally wouldnt add the convienience function (those should be reserved for functions that are heavyly used) but i would extend the call_service.

i could ask you to remove the convienience function and then advise to accept the PR, but the decision about if the function should be added or not i leave to Andrew.

> And you can automate your house without AppDaemon at all, using only HA. We both agree that AppDaemon is great, that's why we use it.
that wasnt my point. (allthough you are right that AD is great ;) ) 
you suggested that AD was only for making automations in HA better/easier/faster, and that if everything that wasnt there for that, AD could be left out completely.

my point is that AD is a program on its own, that has some convienience for HA users. (and everytime something is added, there is the decision to make if its usefull for just 1 person, or for a lot)
but AD isnt (anymore) all about HA. AD can work together with every program that is like HA and hass a rest api. (domoticz, openhab, etc. etc.)

ReneTode:
@dlashua andrew is willing to add the function if you add the docs that need to go with it in the style from the new docs.

you can look at adapi.py from the dev version how to provide the docs.

dlashua:
Documentation for render_template() has been added.

PR #620, opened 2019-08-14
State duration fix

Odianosen25:
This fixes an issue, whereby when the duration is used, and the state changes, it will still execute the previous state timer even when the state had changed. This leads to it being executed multiple times because the previous timers were not properly cancelled.

### DIFF
--- a/appdaemon/threading.py
+++ b/appdaemon/threading.py
@@ -506,6 +506,9 @@ class Threading:
                     if kwargs.get("oneshot", False):
                         kwargs["__handle"] = uuid_
 
+                    if "__duration" in kwargs and kwargs["__duration"] != None: #if a timer running already
+                        await self.AD.sched.cancel_timer(name, kwargs["__duration"])
+
                     kwargs["__duration"] = await self.AD.sched.insert_schedule(
                         name, exec_time, funcref, False, None,
                         __entity=entity,
@@ -533,6 +536,7 @@ class Threading:
                 if "__duration" in kwargs and new != old:
                     # cancel timer
                     await self.AD.sched.cancel_timer(name, kwargs["__duration"])
+                    kwargs["__duration"] = None
 
         return executed
 


ReneTode:
previous timers??
doest it have only one timer, that starts when the asked state is there, and that gets cancelled when the state isnt hold for the asked time?

Odianosen25:
Well I found out it doesn't always cancel the timer in some circumstances. I had noticed it for a while, but as the logic made sense in the code, never got to looking into it.

Now what is going on, is when there is a change, it is supposed to be cancelled as said. But what is going on, is that sometimes the timers are not being cancelled and another is re-initialised leaving the previous one running.

I honestly cannot explain the logic now, about why it happens. But I did put in some logs in my apps and I saw it. With this fix, it always ensured that the duration timer is only executed after the set time has expired. Thereby allowing only a single timer to be executed.

Regards

ReneTode:
i dont see how this can work at all. a summary:
```
if new != old: #  line 500
    if "duration" in kwargs: # 501
        if "__duration" in kwargs and kwargs["__duration"] != None: # 509
              await self.AD.sched.cancel_timer(name, kwargs["__duration"]) # 510
        kwargs["__duration"] = await self.AD.sched.insert_schedule(...) # 512
    if "__duration" in kwargs and new != old: # 536
        await self.AD.sched.cancel_timer(name, kwargs["__duration"]) # 537
```
so in 510 it gets canceled
in 512 a new one started
and in 537 its canceled again, because __duration is always in kwargs at that point and new is always different from old at that point.

Odianosen25:
Well I really get the point you making, but for some reason somewhere line 537 misses cancelling the existing timer.

Now this is how I think it happens, and it centres around now declaring a `new` or `old` when using the duration parameter: 
- When the state changes and `new` or `old` is not specified, the timer is activated and running
- When the state changes again, and `new` is not specified, it will not cancel the previous timer but  run another. As line 500 will be `True`
- This will continue, as long as the state changes, and a `new` is not specified as the code will run is the callback declares no new or old.

Now due to this, if there is a timer running. it will not cancel it and will keep re-initialling a new timer, keeping the handle of only the last one

So what my fix will do now, is to ensure it will not start a new one, unless the previous one is cancelled. This way, multiple will not run

Regards

ReneTode:
i dont think the new and old in that piece of code are the specified new and old, but the actual new and old value.
because else there would be no need to check for new != old. (would be stupid to give a listener for where new == old)

if line 500 is True then new is always != old. 
so line 536 will always be True.
and if line 536 is always True, then the callback that is started in 512 will always be canceled.

so this piece of code can never start a schedular.

Odianosen25:
Oh I think you misunderstood me, apologies I wasn't clearer.

Line 500 is 
```
if (cold is None or cold == old) and (cnew is None or cnew == new) and new != old:
```
Now `cold` and `cnew` are the values declared in the listen_state function. Now if in the listen_state, `new` nor `old` are not given(cnew and cold in this case), then line 500 will be True (as long as `new` != `old`). 

This is what leads to multiple timers being ran, even when there are other ones running already.

Regards

ReneTode:
i  think you dont read what i say.
if line 500 is True then new !=old, right?
keep that in mind.
so the code in the if statement is started.
the code is started so line 512 will be activated. it starts a schedular and saves the value in kwargs["__duration"]

oke, so now a schedular is started and kwargs["__duration"] exist, right.
now look at 536.
it will always be True, because we already checked new != old and we just created "__duration" in kwargs.

that means that 537 will ALWAYS be activated.
what does 537?
it cancels the schedular that we started in 512.

so we ALWAYS end up with NO schedular.

Odianosen25:
Oh ok unfortunately it is not the case, and I explain. Where I noticed this issue prominently, is actually with my `media_player.living_room` and I will use it to explain.

At start up I have
```
self.listen_state(self.media_state_changed, media_player, attribute="source", duration=300)
```
If I am watching "Chromecast" and I switch source to like "Kodi", now as your first point, the timer is started as you rightly said; for `new != old`. Now if I was to switch source again, even if back to "Chromecast", line 500 will be True, since `new != old`. 

it will always be True, as long as I am switching sources, and `cold` and `cnew` are both `None` from my listen function. Since it is not executing line 537, what ends up happening, is that I just get a pilled up number of timers, and later my TV starts behaving weird like its switching sources by itself.  

Hope this explains better?

Odianosen25:
You can also try an experiment, and listen to an input_select. Put in a duration of like 120. Just have a simple log print out when it executes its callback.

Now change the state from the GUI like 5 times or more within 2 minutes. If it works properly, it should only print out once like 2 minutes after the last change. 

But what you will see, is that it will execute the command like 5 times, which is not right. Put in this fix, and try again, and you will see it only runs once.

ReneTode:
dont you understand my logic?

```
if new != old:
  a["b"] = self.run_in(self.something, time)
  if "b" in a and new != old:
    self.cancel_timer(a["b"])
```
is there any way possible that self.something will run in this case?
no self.something wil NEVER run.

i understand that there is a problem. but this piece of code should never work in my opinion.
so what am i overlooking?

Odianosen25:
Trust me boss it works, as I am using it now  ;)

Just try the test I wrote above, and you will see what I am saying.

`self.something` will only run, only if there is no more callback to process, which is what `duration` is there for :)

ReneTode:
would have been helpfull if you read my code and answered my questions. ;)
i just reread the whole code and i missed line 535, so that also was missing in my example.
that was what i was overlooking.

in that case i think your solution is not a solution, but a cover up.
a better way is to make sure that line 538  fires when it should be fired.

Odianosen25:
> would have been helpfull if you read my code and answered my questions. ;)
I read it, but I am honestly having difficulty explaining what is going on to be honest ;). I alluded to that earlier.

> in that case i think your solution is not a solution, but a cover up. a better way is to make sure that line 538 fires when it should be fired.

I agree to that, but as the timer is to be re-initialised immediately in line 509, getting line 538 to cancel might be a tall order.

Regards

ReneTode:
then there is obsolete code in AD, and you dont want that either.

so please create a solution that is 
1) no cover up
2) has no obsolete code.

Odianosen25:
Oh my last response doesn’t attest to the fact it is a cover up, I was referring to the later part of the statement.

I am not sure I can make it nearer than it is, unless Andrew has a better idea on how to handle it

ReneTode:
ill take a closer look later on.
i think its just a matter of adding something to the original if statements.

Odianosen25:
Closing this guys, thanks

PR #621, opened 2019-08-15
MQTT fixes

dlashua:
Two fixes:

1) wildcard topics were matching anywhere in the topic. However, they should only match from the beginning of the topic.  Example:

A Wildcard of `test/#` should match:
test/a
test/b
test/a/b/c/d/e/f
test/test
test/test/a

but it should not match:
a/test
a/test/b
a/b/c/test/d

2) the last line of call_plugin_service() was `return result`. However, in situations like subscribing to a topic that was already subscribed to, no result was generated which led to an error:
```
WARNING AppDaemon: Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/appdaemon/services.py", line 60, in call_service
    return await funcref(ns, domain, service, data)
  File "/usr/local/lib/python3.7/site-packages/appdaemon/plugins/mqtt/mqttplugin.py", line 292, in call_plugin_service
    return result
UnboundLocalError: local variable 'result' referenced before assignment
```

I didn't see any place where `result` was actually being used, so I removed it entirely.

### DIFF
--- a/appdaemon/plugins/mqtt/mqttplugin.py
+++ b/appdaemon/plugins/mqtt/mqttplugin.py
@@ -293,8 +293,6 @@ class MqttPlugin(PluginBase):
             self.logger.warning('Topic not provided for Service Call {!r}.'.format(service))
             raise ValueError("Topic not provided, please provide Topic for Service Call")
 
-        return result
-
     async def process_mqtt_wildcard(self, wildcard):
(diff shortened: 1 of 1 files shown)


ReneTode:
result returns a handle.
if the function gets an error it should return None.

usefull change but i think that at the place where the error is generated result needs to be set to None.

PR #624, opened 2019-08-16
global_modules feature and fix

dlashua:
Features:
This allows global_modules to exist in more than one yaml file. Configuration is merged.

Breaking Change(?):
global_modules must be a list. Warning generated if it isn't.

Bug Fixes:
AppDaemon no longer generates the following error after changing global_modules while running:
```
2019-08-16 12:15:46.008372 INFO AppDaemon: Terminating global_modules
2019-08-16 12:15:46.008994 WARNING global_modules: ------------------------------------------------------------
2019-08-16 12:15:46.009184 WARNING global_modules: Unexpected error initializing app: global_modules:
2019-08-16 12:15:46.009354 WARNING global_modules: ------------------------------------------------------------
2019-08-16 12:15:46.009914 WARNING global_modules: Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/appdaemon/app_management.py", line 752, in check_app_updates
    await self.init_object(app)
  File "/usr/local/lib/python3.7/site-packages/appdaemon/app_management.py", line 222, in init_object
    self.logger.info("Initializing app %s using class %s from module %s", name, app_args["class"], app_args["module"])
TypeError: list indices must be integers or slices, not str
appdaemon-dev      |
2019-08-16 12:15:46.010092 WARNING global_modules: ------------------------------------------------------------
2019-08-16 12:15:46.010860 WARNING AppDaemon: Unable to find module global_modules - initialize() skipped
```

### DIFF
--- a/appdaemon/app_management.py
+++ b/appdaemon/app_management.py
@@ -272,11 +272,13 @@ class AppManagement:
                                 for app in config:
                                     if config[app] is not None:
                                         if app == "global_modules":
-                                            if isinstance(config[app], list):
+                                            if isinstance(config[app], str):
+                                                valid_apps[app] = [config[app]]
+                                            elif isinstance(config[app], list):
                                                 valid_apps[app] = config[app]
                                             else:
                                                 if self.AD.invalid_yaml_warnings:
-                                                    self.logger.warning("global_modules should be a list in File '%s' - ignoring", file)
+                                                    self.logger.warning("global_modules should be a list or a string in File '%s' - ignoring", file)
  
                                         elif "class" in config[app] and "module" in config[app]:
                                             valid_apps[app] = config[app]
--- a/appdaemon/app_management.py
+++ b/appdaemon/app_management.py
@@ -272,7 +272,12 @@ class AppManagement:
                                 for app in config:
                                     if config[app] is not None:
                                         if app == "global_modules":
-                                            valid_apps[app] = config[app]
+                                            if isinstance(config[app], list):
+                                                valid_apps[app] = config[app]
+                                            else:
+                                                if self.AD.invalid_yaml_warnings:
+                                                    self.logger.warning("global_modules should be a list in File '%s' - ignoring", file)
+ 
                                         elif "class" in config[app] and "module" in config[app]:
                                             valid_apps[app] = config[app]
                                         else:
--- a/appdaemon/app_management.py
+++ b/appdaemon/app_management.py
@@ -290,6 +290,11 @@ class AppManagement:
                             if new_config is None:
                                 new_config = {}
                             for app in valid_apps:
+                                if app == "global_modules":
+                                    if app in new_config:
+                                        new_config[app].extend(valid_apps[app])
+                                        continue
+
                                 if app in new_config:
                                     self.logger.warning("File '%s' duplicate app: %s - ignoring", os.path.join(root, file), app)
                                 else:
--- a/appdaemon/app_management.py
+++ b/appdaemon/app_management.py
@@ -402,6 +402,9 @@ class AppManagement:
                 # Check for changes
 
                 for name in self.app_config:
+                    if name == "global_modules":
+                        continue
+
                     if name in new_config:
                         if self.app_config[name] != new_config[name]:
                             # Something changed, clear and reload
@@ -424,6 +427,9 @@ class AppManagement:
                         await self.remove_entity(name)
 
                 for name in new_config:
+                    if name == "global_modules":
+                        continue
+
                     if name not in self.app_config:
                         #
                         # New section added!


ReneTode:
would be acceptable if there was no breaking change.
add a check if a list or a string is given.
when a string is given then load the single file if a list is given then add the files in the list.

dlashua:
> add a check if a list or a string is given.
> when a string is given then load the single file if a list is given then add the files in the list.

This is the part I wasn't sure of.

I didn't modify any of the file loading, file parsing, or config storage code. So these changes only effect the way the config looks internally once it's loaded. Right now, at least in the parts I touched, no manipulation of the config happens at all. It's trivial to check if something is a list OR a string. But the cleanest way to write this is, if a string is supplied, convert it to a list (so that the list can be extended if other yaml files also contain just a string. But I wasn't certain if converting it to a list would break something else somewhere else that handles a string instead of a list in a special way. But as long as nothing is, then converting it to a list will be fine.

ReneTode:
i really wouldnt know if it would cause trouble if it is converted to a list.
but if that is the case then it would also be a problem if you provide a list.

> But I wasn't certain if converting it to a list would break something else somewhere else that handles a string instead of a list in a special way
in that case you are not certain that this PR is not causing any trouble. and that is a problem.
when you want to provide code to AD (which we appreciate) the best way to go is:
1) first talk with us on discord about it. we can point you to possible trouble you run into, or things that you need to think about. maybe its not needed at all and you can do it another way.
2) if you provide code, then at least make sure that you yourself trust that it wont create a problem somewhere. (not 100% sur but 200%)

i also want to point out the following. its nice to be adding stuff in AD itself, but remember that everything you do will be lost when you want to update AD. and there are big changes and great additions coming in the next version(s)
so its wise to help yourself by only changing those things that have a chance to be implemented. (and those chances increase when you talk before you act)
and if your not on hassio, you can also use our dev version, which would keep you up to date and when a PR from you is added it would be integrated in your system directly.

so when it comes to this case:
please try to find out if converting to a list could cause trouble anywhere else. if you are absolutely convinced that wouldnt cause any trouble then change the code so that it wont give any breaking changes.

i hope to see you on discord, about this or other subjects.

dlashua:
> i really wouldnt know if it would cause trouble if it is converted to a list.
but if that is the case then it would also be a problem if you provide a list.

This was my assumption too.


> and if your not on hassio, you can also use our dev version

I am running the current "dev" tree, self identifying as 4.0.0b1. I couldn't find a docker image for the current tree, so I built one available at dlashua/appdaemon:dev.

I have a testbed running as well which incorporates any PRs I've submitted so I can make sure the code I've sent isn't bad. It's available at dlashua/appdaemon:dev-new


> so its wise to help yourself by only changing those things that have a chance to be implemented. 

I like to stay with the mainstream whenever possible. For bug fixes, I just fix them and PR, because they are bugs. For features, I try to implement it with the tools available (like the "app" I've written that serves to monitor the imports an app uses and restart those apps when the underlying module changes without having to specify "global_dependencies" on every single use of the app). But, when it's not possible, or really hacky, or a feature I think other people will want/need, then I make the change in AD. If the PR isn't accepted and the concept of the PR is flat out rejected then I have to decide if I want to handle it the hacky way or divert from mainstream AD.

To date, I've submitted 4 PRs to AppDaemon. One fixes an oversight in the Dockerfile and was accepted almost immediately. Another fixes a bug in the mqtt plugin (that caused my AppDaemon instance to enter and infinite loop of mqtt publishes) and cleans up some error handling. Another fixes an error dealing with global_modules and adds a small feature that allows easier sharing of AppDaemon apps. And the last one simply adds a feature (hass template rendering). I agree, that last one is not of great use to lots of people, but it was the only API feature Home Assistant provides that there isn't a service call in AD for, I had a need for it, and implementing it without changing the hass plugin would have been tedious.

> first talk with us on discord about it.

I tried this. There are very few people actively on the AppDaemon discord except for you. 

> if you provide code, then at least make sure that you yourself trust that it wont create a problem somewhere. (not 100% sur but 200%)

I can't be 200% sure of anything. I'm certainly not as well versed in the AppDaemon core as Andrew is. But this is why it's a "Pull Request" and not a "Pull Demand"/"Force Push". So that someone more knowledgeable can make certain the changes aren't breaking something else somewhere else. To my knowledge, these PRs don't cause any issues and, at least for me, they are running and working with no issues.

ReneTode:
> I tried this. There are very few people actively on the AppDaemon discord except for you.
if you got my attention then thats enough ;) if i cant speak for the whole team i let you know, but ill always make sure that it gets Andrews attention when needed. Odia reads most of it also.

> I can't be 200% sure of anything. I'm certainly not as well versed in the AppDaemon core as Andrew is. But this is why it's a "Pull Request" and not a "Pull Demand"/"Force Push". So that someone more knowledgeable can make certain the changes aren't breaking something else somewhere else.
when it comes to the core then Andrew is hard to beat. but still he needs to take time to check a PR.
and when the person that offers the PR, isnt 200% convinced that there is no problem, he needs to triple check instead of double check, and with his little time lately that means that those things will be left on the side. thats why you need to be as convinced as possible and in the description show that you looked deeper and tested thorowly. everything that helps Andrew to use less time deciding is important in such cases. (thats also where my opinion comes in ;) )

im glad to see that you are on dev, that helps a lot in such cases.

about the template rendering, (you mention it here, so ill respond) split the PR into 2 parts.
1 that provides the convienience function and 1 that makes the services available, and ill ask Andrew to add the last 1 immediatly, because i think its valuable to provide all services that HA provides.

i really look forward to talk to you on discord about your next possible PR ;) or anything else.

PR #627, opened 2019-08-19
 Add call_service documentation and examples

rhumbertgz:
This documentation is based on the one found it on Hass API reference

### DIFF
--- a/appdaemon/adapi.py
+++ b/appdaemon/adapi.py
@@ -1314,13 +1314,35 @@ class ADAPI:
         self.AD.services.register_service(namespace, d, s, cb)
 
     def call_service(self, service, **kwargs):
-        """
+        """Calls a HASS service within AppDaemon.
+
+        This function can call any service and provide any required parameters.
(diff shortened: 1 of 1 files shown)


ReneTode:
@rhumbertgz this description is about the call_service from hass api.
but you placed it with the call_service from AD api.

there are also other services that can be called besides the HA services.

rhumbertgz:
@ReneTode I though that hassapi.py extends adapi.py, and hassapi.py didn't change the implementation of the parent I should put the description on the original code (Parent).

ReneTode:
adapi is the general. it provides functions that should work with all plugins and also has functions that dont need a plugin at all.
hassapi contains api specific for the hass plugin. it will have functions that are not in adapi or any other plugin, but it can also have functions that are in adapi, but that are plugin specific.

in this case the call_service originally only existed in hassapi, but we extended it to be able to use AD internal services. and possibly in the future other plugins can use the same.

PR #635, opened 2019-08-23
Update call_service examples

rhumbertgz:
Examples provided by @Odianosen25

### DIFF
--- a/appdaemon/adapi.py
+++ b/appdaemon/adapi.py
@@ -1341,11 +1341,25 @@ class ADAPI:
             None.
 
         Examples:
+            HASS
+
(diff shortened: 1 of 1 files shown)


Odianosen25:
@rhumbertgz thanks for mentioning me

on this side of things
```
For Utility, it is important that the `namespace` arg is se to ``appdaemon`` as no app can work within that `namespace`. If not namespace is specified, calling this function will rise an error.
```

it should be `set` not `se` ;)

rhumbertgz:
> it should be `set` not `se` ;)

Fixed it!

PR #640, opened 2019-08-27
Update MQTT API Reference Documentation

rhumbertgz:
@Odianosen25 Please check the changes made to the MQTT_API_REFERENCE.rst

### DIFF
--- a/appdaemon/adapi.py
+++ b/appdaemon/adapi.py
@@ -90,6 +90,7 @@ class ADAPI:
 
         Args:
             msg (str): The message to log.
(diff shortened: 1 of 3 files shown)


Odianosen25:
Hello @rhumbertgz, thanks for this. But just the `listen_event`. It states "Function to be invoked when the requested state change occurs". It should be more like "Function to be invoked when the requested event occurs".

Kind regards

rhumbertgz:
Hi @Odianosen25 

I updated the description :)

> It should be more like "Function to be invoked when the requested event occurs".

PR #645, opened 2019-08-28
WIP: ASYNC for app modules, call backs, and init/term

dlashua:


dlashua:
Don't merge this yet. There's a bug in the changes to threading.py. Waiting on @Odianosen25 for a fix.

dlashua:
The code to handle async callbacks has been removed. It will come in a separate PR. This PR just handles arbitrary async calls from inside an app module.

Issue #676, opened 2019-09-15
Feature Request: App-local data storage


efficiosoft:
Hi,

I'm not sure whether this is planned for 4.0 and, even if so, whether you already implemented it, but I'd like to have an uniform way to store app-specific data, probably best as a persistent key-value store.

If desired, I could provide a thread-safe implementation using Python's `shelve` with an API like this:

```
self.open_storage()  # When called the first time, a persistent db file is created based on app name
self.store["foo"] = [1,2,3]  # writes synchronously to the file, using the regular dict API
```

On app shutdown, AD would just close the store if it has been opened, no invasive code changes needed.

What do you think?

Best regards
Robert

Odianosen25:
@efficiosoft,

This has already been implemented in AD 4.0. AD now has the ability to define [user defined namespaces](https://github.com/home-assistant/appdaemon/blob/4.0.0b1/docs/CONFIGURE.rst#advanced-appdaemon-configuration). These are namespaces, not tied to any plugin and can be used as desired. it can also be set such that when data is stored there, it saves into file the present state (depending on the performance required). 

These namespaces can also be persistent, and so all entities stored in here, are loaded up at boot time. I use this, in conjunction with an app, to get data I want stored or state of some entities I need, in here, and at start up, these entities are loaded back into the required namespaces they are to operate. For example windows or door sensors.

Regards

efficiosoft:
@Odianosen25 Interesting, thanks. I guess the entities are serialized as JSON?

Odianosen25:
You welcome @efficiosoft,

Yes they are. So data that has issues with serialisation cannot be stored in there.

Odianosen25:
Please close the issue when you satisfied.

Thanks

ReneTode:
@efficiosoft i also found this:
https://appdaemon.readthedocs.io/en/dev/AD_API_REFERENCE.html#appdaemon.adapi.ADAPI.save_namespace

and you might not know it but it can be helpfull for you, we now got an index in the docs:
https://appdaemon.readthedocs.io/en/dev/AD_INDEX.html#s

efficiosoft:
Ok, I still think it would be useful to have a kind of implicit per-app namespace that can be used to store internal state, especially when using third-party apps as-is. Do you think that's feasible?

efficiosoft:
@ReneTode Thanks for the links. I indeed only saw the stable versions built on rtfd.

ReneTode:
[...]

PR #677, opened 2019-09-15
Use shelve to persist user-specific namespaces

efficiosoft:
Related to #676.

This changes `PersistentDict` to use shelve as backend. We don't need locking anymore, verified at least with Berkeley and GNU dbm.

### DIFF
--- a/appdaemon/state.py
+++ b/appdaemon/state.py
@@ -26,14 +26,8 @@ class State:
                 os.makedirs(nspath)
             for ns in self.AD.namespaces:
                 self.logger.info("User Defined Namespace '%s' initialized", ns)
-                writeback = "safe"
-                if "writeback" in self.AD.namespaces[ns]:
-                    writeback = self.AD.namespaces[ns]["writeback"]
-
-                safe = False
-                if writeback == "safe":
-                    safe = True
-
+                writeback = self.AD.namespaces[ns].get("writeback", "safe")
+                safe = bool(writeback == "safe")
                 self.state[ns] = utils.PersistentDict(os.path.join(nspath, ns), safe)
         except:
                 self.logger.warning('-' * 60)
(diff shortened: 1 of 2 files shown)


efficiosoft:
I was wrong, there's a possible race condition with `writeback` in `sync()` which is now resolved. All potential interactions with the underlying backend are now serialized. My tests showed `dbm.ndbm` to be multiple order of magnitudes faster than `dbm.gdbm` with the safe writeback strategy, which should be fine since it's the default on systems with no gdbm installed.

acockburn:
Thanks Robert - I'll take a look at this probably over the weekend with a view to getting it in the next Beta.

Odianosen25:
Hello @efficiosoft,

Thanks for this, just something I want to understand if you don't mind.

- Is it possible to delete an existing persistent namespace using this method, and has it been built in? Reason asking is, if we are going to give the ability to create namespaces via apps, there is also the need to be able to delete as needed via app. I have gone via the code, and I am not certain this is in. I know there is the `delitem()`, but that's not deleting it as a whole. As simply `self.state.pop(ns)` will still most-likely keep the previous data in the database.

- Also though this might answer the first one, when the `self.state[ns].sync()` is used, if a key had been deleted from the namespace before, does it remove none existing keys, at the same time as add new ones?

efficiosoft:
Hi @Odianosen25 

Regarding your first question, this is not part of the actual storage implementation and this PR doesn't add anything in this regard. Namespace deletion will have to be handled externally, I don't know what the planned API for this will look like.

`sync()` persists any changes that were made to the namespace since the last time `sync()` was called, so yes, it also commits key deletions.

Best regard
Robert

Odianosen25:
Hello @efficiosoft,

Thanks for your response.

> Regarding your first question, this is not part of the actual storage implementation and this PR doesn't add anything in this regard.

Yes that is true, but since its been asked that namespaces be created via api calls, it became needed; so wondered how it will work using shelve. When you say externally, can you kindly give an example?

Or the database will need to be opened, and that deleted or something? Is it something you could also add, so when building the namespace apis, can easily be built in?

acockburn:
@Odianosen25 - this should be close to a drop in replacement for what we already have. Creating a namespaces is really just instantiating a persistent dict as it was before, but with a different implementation under the covers. If we need to add a method to close and delete the DB we will do so - I don;t think Robert needs to change anything at this point.

efficiosoft:
@Odianosen25 Ah, now I see what you're after. These databases are just plain files, one per namespace. So deleting them is a matter of `ns.close()` and removing the file thereafter, both of which could easily be done from the `delete_namespace()` function (or however that's going to be named).

PR #682, opened 2019-09-17
Move log to Admin Namespace

Odianosen25:
These changes proposes the move of the `_AD_LOG_EVENT`, from the `global` to the `admin` namespace.

Being in global, when having other log events from other namespaces (AD Plugin), it is not possible to subscribe to only remote log events, without getting the local log events also. 

Also being log, might actually be better being kept in the Admin, so if there is need to develop some restrictions to Admin rights, it becomes easier, since log is an `admin` thing.

Regards


acockburn:
Not sure about this - this removes the restriction for having events in admin which was put in place to avoid many nasty things form happening ...

Odianosen25:
Hmmm

The restrictions we did put in place then for Admin, as to avoid the ability to listen to `state_changed` on a few entities, but not exactly events in general.

So that was why in this [line](https://github.com/home-assistant/appdaemon/blob/a9dbe4efc0632a2481313f6c1b330cdb58b1b82d/appdaemon/events.py#L154) I still left in.

Outside that, I don't remember anything unless I am not aware of it.

acockburn:
OK, I'll merge it - but if we start getting bad things happen I'll make you fix it ;)

Odianosen25:
No problem boss

PR #701, opened 2019-09-20
Remove unnecessary code

rhumbertgz:


Odianosen25:
Hello @rhumbertgz,

Thanks for this, but I don’t think this will be accepted. Reason being that state callbacks, are only for “state_changed” events and not for all events. 

If this is put in, it will mean state callbacks being ran for each event type, which is unneeded and extra processor usage which is unnecessary. 

Regards

rhumbertgz:
> If this is put in, it will mean state callbacks being ran for each event type, which is unneeded and extra processor usage which is unnecessary.

@Odianosen25  look at line 169, the `if` condition of that line also check  `if data['event_type'] == "state_changed":`, for that reason the inner `if` condition is unnecessary

Odianosen25:
Oh that's a good point again.

Thanks

PR #704, opened 2019-09-20
Add attributes declared outside the `__init__` method to it.

rhumbertgz:
I found several attributes declared outside the `__init__` method. 
The best practice is to declare attributes in `__init__` before using them in other methods.
I just added those attributes to the `__init__` method.

### DIFF
--- a/appdaemon/app_management.py
+++ b/appdaemon/app_management.py
@@ -24,8 +24,8 @@ class AppManagement:
         self.monitored_files = {}
         self.filter_files = {}
         self.modules = {}
-
         self.objects = {}
+        self.check_app_updates_profile_stats = None
 
         # Initialize config file tracking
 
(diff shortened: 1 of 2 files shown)


Odianosen25:
Thanks for this @rhumbertgz,

But those variables, are not used across the class but only within certain functions. For that reason, not sure there is need to have them declared across the class.

rhumbertgz:
> Thanks for this @rhumbertgz,
> 
> But those variables, are not used across the class but only within certain functions. For that reason, not sure there is need to have them declared across the class.

hi @Odianosen25, based on your comment they should be **local** variables instead of **global** ones in my opinion  🙂

Odianosen25:
Oh apologies and you right. I forgot Andrew made a change to the plugin, which requested those variables to be global. So yeah you got a good point.

Regards

PR #710, opened 2019-09-20
Simplify boolean conditions

rhumbertgz:


appdaemon/plugins/mqtt/mqttplugin.py
```diff
@@ -401,7 +401,7 @@ def start_mqtt_service(self, first_time):
                     auth.update({"keyfile" : self.mqtt_client_tls_client_key})
                     set_tls = True
                    
-                if set_tls == True:                    
+                if set_tls:
```
acockburn: For this I prefer `if set_tls is True:`

appdaemon/scheduler.py
```diff
@@ -381,7 +381,7 @@ def get_next_entries(self):
                 if self.realtime is True:
                     self.now = now
                 else:
-                    if result == True:
+                    if result:
```
acockburn: and this

appdaemon/plugins/mqtt/mqttplugin.py
rhumbertgz: done!

appdaemon/scheduler.py
rhumbertgz: done!

PR #711, opened 2019-09-20
use __name in add_event_callback()

dlashua:


acockburn:
Looking at this again, I'm not sure what the issue was - `_name` is purely a variable internal to `add_event_callback()`, it doesn;t end up in the kwargs anywhere so it should be a benign change to rename it, so we can leave it as is and benefit from slightly improved docs readability

acockburn:
Notes: Adding name to add_sevent_)callback() break anything with a kwarg of name
We may just mark this as a breaking change

acockburn:
OK, from the users perspective, `name` is a hidden undocumented variable we are passing through on their behalf so it's not fair to break their code with this.

PR #737, opened 2019-09-27
Mqtt API async fix

Odianosen25:
This puts the MQTT API in line with the present async framework


acockburn:
This PR has a lot of stuff in it completely unrelated to MQTT - I'm not happy about merging it as is. Looks  like it has some additional stuff for the AD plugin, it adds fire_event() it's too much to digest without knowing exactly what is going on.

Odianosen25:
Apologies wasa mistake

PR #744, opened 2019-10-01
add sleep to adapi

dlashua:


appdaemon/adapi.py
```diff
@@ -2711,4 +2711,22 @@ def callback_inner(f):
...
+
+        Returns:
+            result or None.
+
+        Examples:
+            >>> async def myfunction(self):
+            >>>     await self.sleep(5)
+        """
+        
+        return await asyncio.sleep(delay, result=result)
```
acockburn: Daniel, could we add an error check to log a message if an attempt is made to call this from a thread? Alternatively, what would the untrapped  error look like? would it be easy to spot, or would it cause weird async things to happen ??

dlashua: @acockburn I can add a check and raise an error if you'd like.

Without a check, calling adapi.sleep() would result in the return of a coroutine. Without jumping through a lot of hoops, a sync method would have no way of "awaiting" that coroutine, in which case, it would do nothing, including not sleep. The same thing it would do if called in an async method without "awaiting" it in some way.

I looked at ways of doing a "sleep" in a sync method while still allowing the thread to perform other callbacks and such. Without really weird syntax (that can already be done using run_in()) there is no way. The threads can only do one thing at a time. If it is sleeping, it can't do anything else.

Alternatively, I could make this call time.sleep() in a sync thread, which would, in fact sleep. But, as you know, it would also delay all other callbacks in that thread during the sleep.

acockburn: OK, then I think to avoid confusion with  users of threads who don't necessarily understand async we should just raise an error - thanks!

dlashua: I have added the Exception for sync apps.

PR #746, opened 2019-10-02
Remove unnecessary "--virtual" from Dockerfile

benleb:
`--virtual` is used to create a virtual group of the packages in the command (to remove them at once for example (see here https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md#virtual-packages))

as we dont remove the build-deps anymore, giving them a name is useless


acockburn:
Thanks for spotting this!

PR #750, opened 2019-10-05
Update materialdesignicons to 4.4.95

clayhill:
The svg webfont was removed because it is not part of the official github repo anymore.


acockburn:
Thanks!

PR #781, opened 2019-10-26
use en_US form of canceled

benleb:
see https://www.grammarly.com/blog/canceled-vs-cancelled/

US is preferred over UK, isnt it?


acockburn:
Not a safe assumption considering I am actually British by birth ...

benleb:
ah nice to know :D okay so, would you prefer british? how to decide this? :D

acockburn:
Was just joking - I usually try to keep things consistently US but that one obviously crept through!

PR #792, opened 2019-11-01
Call Service update

Odianosen25:
- Added `call_service` event, when a service is called
- Made changes to the api docs on `call_service`


acockburn:
Won't this result in duplicate events when used in conjunction with HASS? Seems like this should only be done for services that belong to HASS and not a plugin. The plugin should be responsible for either propagating any service call events back to AD (e.g. in the case of HASS< which we get from it's event stream), or for creating them itself ehrn the service is called.

Odianosen25:
Well I understand, it was a request from @ReneTode, so will amend it as the docs still need working on, will ask Rene what best works for him

ReneTode:
hmm, hass triggers indeed an event when a service is called.

so we would need to differentiate. i think its logical to have an event when a service is called (like HA does) so there should be an option inside plugins, that if they have service events, AD wont generate 1.
or we only generate service events for AD services (but how to know the difference between what is a plugin service or an AD service?)

or all service calls generate an event in an AD namespace (global?)
and you have the event duplicated in 2 different namespaces, so you can decide where to listen and just get 1

Odianosen25:
Instead of sending it to global, which will still duplicate, we could send it to `admin` or best `appdamon` namespace. This way anyone that needs it, gets it from there.

This will avoid duplicate within the same namespace. I said `appdaemon` because `admin` has `fire_event` service call within it, and that will mean double events when `fire_event` is called. So since `appdaemon` doesn't have it, can be placed there

PR #798, opened 2019-11-03
Support for HA api tokens required to get camera images

rds76:


ReneTode:
AD 4 beta already got a camera widget that doesnt need to set any token.
and the docs for this are edited in the iframe section, not the camera section.

rds76:
Just checked. Good idea to read camera tokens from their attribute.
So far, it is fix for those running current version. And mabe it could be usefull not just for HA camera, but as universal solution for url call with Bearer auth token.

ReneTode:
if you want to call an url then you can use the iframe widget.

those who can fix it themselves can also download the camera widget from the dev version.

rds76:
Yes, thats what I did. Already backported basecamera to v3 from dev.
Also made few improvement in .js to make it work even better. [Check it out](https://github.com/home-assistant/appdaemon/compare/dev...rds76:dev)
What I also did not like that basecamera is clickable with <A> tag. When you run dash in kiosk mode, you cannot go back to dash then.

ReneTode:
making camera clickable was on the wishlist from a lot of users.
but it should be optional.
which it is going to be when we completely renew dashboard after AD 4 is released.

what are the improvements you are referring to and what do they adchieve?

rds76:
Guess clickable - they asked for fullscreen view - check my fullscreen [hack](https://gist.github.com/rds76/496459db4575c533f64aff2b611ea069) I made in the past.

Improvements: 
- I exchanged ugly white gif background for transparent as before dash gets access token there was noticable blink effect on dash load
- made stream option 'exclusive' when set together with refresh

ReneTode:
there are 2 groups of people. 
the first group likes just the cam fullscreen, the other group wants to navigate to a page that has a bigger view from the cam.
nice hack, but a skin is not the right place for such options when we include it.

refresh should also work when stream is set. so thats no improvement but the opposite.

rds76:
what is the sense setting refresh for stream? especially when it is refreshed by change of acces_tokem...

ReneTode:
some people have cams from which the stream isnt that dependable, or the browser doesnt handle cashing as well as it can be. or HA just stops feeding the stream.

sometimes 1 of my cams just stops, and seems to hang for example.
with a refresh the stream is then reconnected.

rds76:
ok, make sense, but change of token is quite frequent so it gets refreshed anyway
at least, plz fix white blink effect ;)

ReneTode:
please open a new issue or PR for that change only.

if you make a PR for that change only i will ask Andrew to include it ASP.
else it will be included in AD 5.

acockburn:
We aren't adding new functionality to AD3 pending the release of AD 4, which already has a fix for the camera, so closing.

PR #805, opened 2019-11-12
fix code style in examples

basnijholt:
I used pyupgrade, isort, and black to fix the style, so this shouldn't result in any breaking changes.

In this way, the code style in `appdaemon` is consistent with `home-assistant`.

### DIFF
--- a/conf/apps/hello.py
+++ b/conf/apps/hello.py
@@ -1,12 +1,13 @@
 import hassapi as hass
+
 #
 # Hello World App
 #
 # Args:
 #
 
-class HelloWorld(hass.Hass):
 
-  def initialize(self):
-     self.log("Hello from AppDaemon")
-     self.log("You are now ready to run Apps!") 
+class HelloWorld(hass.Hass):
+    def initialize(self):
+        self.log("Hello from AppDaemon")
+        self.log("You are now ready to run Apps!")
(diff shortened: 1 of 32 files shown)


ReneTode:
> In this way, the code style in appdaemon is consistent with home-assistant.
why should we want that?

if it was to make it consistent with any python style convention i could see it.
but AD doesnt want to be like home-assistant.

basnijholt:
Let's say PEP8 then 😄

ReneTode:
i see changes that effect the code, so thats not good.
and i see no reason why hassapi should be imported after all others.
i always import hassapi first, because its the only line that is common in all my apps.

basnijholt:
PEP8 Suggests to put imports from stdlib first, then external packages, and then finally your own defined modules.

Additionally, it suggests to sort these alphabetically in each category.

ReneTode:
i must correct that. i missed 2 lines (in 1 app) and thought they were removed. but because of the way github presents the changes it only looked like they were removed.

hmm, in my eyes complete useless changes, but i leave it to Andrew to judge it. ;)

basnijholt:
I have split it up into 3 separate commits, if the import sorting isn't needed for example, I can just delete that commit.

dlashua:
I appreciate code consistency. And using "black" is certainly one of the easiest ways to get there which is why I use it in code that I write.

This is a large PR, though. So it'll take some time to go over it all and ensure nothing bad has happened. And github isn't making it any easier by the way it is presenting the diff.

As @ReneTode said, ultimately, code style conventions are up to @acockburn. However, assuming no changes other than style have been made accidentally, I think this is a step in the right direction.

acockburn:
CLosing this as we now have black and flake on the case

PR #806, opened 2019-11-13
Update futures.py

Odianosen25:
Fix for an error, if the user was to cancel a future, and the app terminates and AD tries to cancel the same future


dlashua:
In my testing, a future can be in only 2 states: running, or done.

Once a future is "done", several flags are possible: cancelled, and exception.

So a cancelled future is also done. And a future with an exception is also done.

So I don't think this PR actually changes anything. However, I don't think it hurts anything either.

PR #808, opened 2019-11-13
simplify if-statement

basnijholt:


ReneTode:
is this usable with python 3.6?
and i must say that if it doesnt bring any advantage i would rather not see such chages, because they are less readable.

basnijholt:
I would say this is **much** more readable! `pop` has existed since the beginning of Python basically.

ReneTode:
i never used pop, and its only readable if you know what pop does, which isnt something that is available in other languages.

i dont know pop, i never used pop, so for me it makes it less readable (and so also for every python newby ;) )

dlashua:
pop() is used in many places in the AD code base. In some cases, pop() is much faster than other methods. And, at least for me, pop() is much more readable.

The only thing to consider here is that pop() will remove the key from the dict, which was not happening before this PR. So we need to be certain that we want that data pulled from the dict and, as long as that's the case, this is cleaner IMO.

basnijholt:
@dlashua, oops, my bad. It should be `get` to make it identical.

appdaemon/state.py
```diff
@@ -101,17 +101,16 @@ def terminate(self):
             # In the case of a quick_start parameter,
             # start the clock immediately if the device is already in the new state
             #
-            if "immediate" in kwargs and kwargs["immediate"] is True:
-                __duration = 0 # run it immediately
+            if kwargs.get("immediate"):
```
Odianosen25: @basnijholt this area will lead to a break in code, since immediate is not always supplied. so should be kwargs.get("immediate", False). But the issue with even using this method, is the user could supply any string
and it will still go thru. so should be 
```
if kwargs.get("immediate", False) is True
```

appdaemon/state.py
```diff
@@ -101,17 +101,16 @@ def terminate(self):
...
                 __attribute = None
                 run = False
-                
+
                 if entity is not None and entity in self.state[namespace]:
                     run = True
-                    
-                    if "attribute" in kwargs:
-                        __attribute = kwargs["attribute"]
+                    __attribute = kwargs.get("attribute")
```
Odianosen25: Same thing as before, `attribute` is not always supplied, so should be `__attribute = kwargs.get("attribute", None)`

Odianosen25:
@basnijholt I added some comments in your PR, kindly review.

Thanks

appdaemon/state.py
basnijholt: In the docs it says immediate is a bool and by default `get` returns `None`, and `bool(None) == True`, so this is exactly the same.

One should never do `if attr is True` **unless** the variable can be different types, like integers, but this is not the case here.

appdaemon/state.py
basnijholt: `kwargs.get("attribute", None)` and `kwargs.get("attribute")` both return `None`, that is the default value of e.g., `dict.pop` and `dict.get`.

ReneTode:
> One should never do if attr is True unless the variable can be different types, like integers, but this is not the case here.

i think thats exactly the case here.
the user SHOULD give a bool, but he could also give "true" or "on"or 1
and if the user gives that it should be False.

basnijholt:
@ReneTode, good point.

I think that in that case, it is probably better to explicitly check the type that the user passes, because if he or she uses `"on"` it will seem like it works, even though the opposite happens.

ReneTode:
with he original code:
```
if "immediate" in kwargs and kwargs["immediate"] is True:
```
it wont work for
immediate = "on"

unless i understand something wrong.

and maybe now you understand why i am not a fan from changing correct working code.
2 people are wasting time to check if this code is correct, and discussing about it, and when its accepted in the end (which needs a review from the 3th and last person), AD doesnt have any more functionality then before.

appdaemon/state.py
dlashua: @basnijholt I think you mean `bool(None) != True`, right?

What we want here is if a user supplied `True` for immediate, then the immediate happens. If the user supplies nothing, or anything other than `True` (including `"yes"` and `"True"`) then the immediate should not happen.

So, I believe the most concise way to write this is a combination of the @basnijholt  and @Odianosen25 code:
```
if kwargs.get("immediate") is True:
```

appdaemon/state.py
dlashua: In this case, None is the default return from `get()` if the key is not present. So:
```
kwargs.get("attribute")
```
and
```
kwargs.get('attribute', None)
```
are exactly the same.

appdaemon/state.py
basnijholt: Yes, I meant `bool(None) != True`.

If that behaviour is required then the `... is True` is indeed the correct solution.

But wouldn't it be a better user-experience if we explicitly check the input types? Because if someone writes `"on"`, a user might expect that the code works (as no error is raised.)

dlashua: It would be a better user-experience, absolutely, if input types were checked and an error raised if the input does not validate. I was merely aiming for "same functionality as before" with this.

@acockburn Agrees that input validation is an eventual goal and has mentioned Voluptuous a few times as a method of getting there. 

For many Python APIs, Voluptuous is overkill on a simple method call because Python programmers can generally work this out on their own. But, in the case of AppDaemon, we are encouraging VERY NEW programmers to write Python. Anything we can do to provide a clear error is helpful to them. And a user writing immediate="True" instead of immediate=True is a fairly common mistake.

basnijholt: What about validating the types based on type annotations and using the following decorator:
```python
from functools import wraps
from inspect import getfullargspec
from typing import get_type_hints

def validate_input(obj, **kwargs):
    for name, type_ in get_type_hints(obj).items():
        if name == 'return':
            continue

        if not isinstance(kwargs[name], type_):
            raise TypeError(
                f'Argument {name} is not of type {type}'
            )


def type_check(f):
    @wraps(f)
    def wrapped_decorator(*args, **kwargs):
        # convert *args into **kwargs
        func_args = getfullargspec(f)[0]
        kwargs.update(dict(zip(func_args, args)))

        validate_input(f, **kwargs)
        return f(**kwargs)

    return wrapped_decorator
```

basnijholt: Oh, in this case, that would quite work actually, as the function uses `kwargs`.

PR #809, opened 2019-11-14
Replace Attributes fix

Odianosen25:
If the replace flag was used in `set_state` without `attributes`, it didn't work properly.

This fix ensures as long as the replace flag is used, it will always replace the attributes supplied


ReneTode:
since when is there a replace flag?
why is there no mention from it anywhere?
and what does it do? because normally a state always replaces the old state.

dlashua:
@Odianosen25 something about this doesn't look right. I assume it's supposed to look like this when used:

```
self.set_state('thing.thing', state="on", attributes={"a":"b"}, replace=True)
```

However, the code flow in this PR only looks for "replace" if there are no "attributes".

Odianosen25:
Yes that's right and I will be changing it

dlashua:
Ah perfect! This looks good now!

PR #816, opened 2019-11-20
Added oneshot to event listening

### LINKED ISSUE #813: add oneshot to listen_event
I noticed that `listen_state` has `oneshot` which seems very useful. I think it makes sense to implement the same in `listen_event`. What do you guys think?

I might give this a shot if I have time soon.

Odianosen25:
This adds the ability to fire the callback for a listen event only once

This fixes #813

### DIFF
--- a/appdaemon/events.py
+++ b/appdaemon/events.py
@@ -259,6 +259,7 @@ class Events:
                 self.logger.debug("Discarding event for loop avoidance")
                 return
 
+        removes = []
         for name in self.AD.callbacks.callbacks.keys():
             for uuid_ in self.AD.callbacks.callbacks[name]:
                 callback = self.AD.callbacks.callbacks[name][uuid_]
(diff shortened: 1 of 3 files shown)


appdaemon/events.py
```diff
@@ -300,6 +301,15 @@ def has_log_callback(self, name):
...
                                                                             "kwargs": callback["kwargs"]
                                                                         })
+                                
+                                # Remove the callback if appropriate
+                                if executed is True:
+                                    remove = callback["kwargs"].get("oneshot", False)
+                                    if remove is True:
+                                        removes.append({"name": name, "uuid": uuid_})
+
+        for remove in removes:
```
basnijholt: Take care that you are overwriting `remove`. It is probably better to use a different variable name.

Odianosen25: Well I could, but `remove` is really of no serious use, so I could use it.

Works anyway ;)

basnijholt: What do you mean, no serious use?

PR #818, opened 2019-11-28
Added List services API

Odianosen25:
This adds the ability for apps to request services via API calls, based on namespaces


acockburn:
You need ot add this explicitly into the reference docs, it doesn't get added automatically.

Odianosen25:
Oh ok will see how that works and amend

PR #820, opened 2019-12-04
Add Azure Pipelines and pre-commit linters

### LINKED ISSUE #780: python-dateutil dependency issue on install
Hello, thank you for providing appdaemon.

I thought I'd try setting up a new venv from the dev branch. Unfortunately its broken at the moment:

```
% git clone -b dev https://github.com/home-assistant/appdaemon.git                                                                                      Cloning into 'appdaemon'...
% cd appdaemon                                                                                                                                          

~/dev/appdaemon-dateutil/appdaemon • python3 -m venv env                                                                                                                                     ~/dev/appdaemon-dateutil/appdaemon • source env/bin/activate                                                                                                                                 ~/dev/appdaemon-dateutil/appdaemon • pip install -e .                                                                                                                                        Obtaining file:///home/chris/dev/appdaemon-dateutil/appdaemon
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/home/chris/dev/appdaemon-dateutil/appdaemon/setup.py", line 6, in <module>
        from appdaemon.utils import (__version__)
      File "/home/chris/dev/appdaemon-dateutil/appdaemon/appdaemon/utils.py", line 13, in <module>
        import dateutil.parser
    ImportError: No module named 'dateutil'
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /home/chris/dev/appdaemon-dateutil/appdaemon/

```
The problem is due to `setup.py` importing `utils.py` in order to get at `__ version __`. The new version adds a load of other utility functions, one of which needs `python-dateutil`.

The problem can be worked around by moving the import in question into the function that actually uses it:

```
diff --git a/appdaemon/utils.py b/appdaemon/utils.py
index 83fc8be..28c52c8 100644
--- a/appdaemon/utils.py
+++ b/appdaemon/utils.py
@@ -10,7 +10,6 @@ import pstats
 import shelve
 import threading
 import datetime
-import dateutil.parser
 import copy
 import json
 from functools import wraps
@@ -446,6 +445,7 @@ def check_path(type, logger, inpath, pathtype="directory", permissions=None):
 
 
 def str_to_dt(time):
+    import dateutil.parser
     return dateutil.parser.parse(time)
 
```
 
I think, though, that it would be safer to move that version string out of utils.py to a place that can be imported alone by `setup.py` without needing to import the rest of `utils.py`.

It would probably make sense to adjust the users of __version__ to point to a new place, meaning the patch would be bigger. I didn't want to start on doing that without any feedback from yourselves.

Another solution would be to place the version in a plaintext file that could be read by `setup.py` and `utils.py` in the same way as the Python packaging guide recommends reading README.md in https://packaging.python.org/guides/making-a-pypi-friendly-readme/

```
this_directory = path.abspath(path.dirname(__file__))
with open(path.join(this_directory, 'README.md'), encoding='utf-8') as f:
    long_description = f.read()
```

basnijholt:
This PR adds Azure Pipelines CI that automatically checks code style using `black` and finds bugs from analyzing the code using `flake8`.

For now, the only thing that happens are those things, however, it is easily extendible in the future to include unit tests that run on different operating systems, different Python versions, etc. I didn't do that yet, because I don't want to make too many changes at once.

I add `pre-commit`, which is a framework that handles git pre commit hooks. It is really easy to use, you just do `pre-commit install` **once** and then each time you make a commit it will perform the checks. If a user didn't do this, then the CI will catch the errors.

When merged, @acockburn should coordinate with me on chat and I will give you full access of the `appdaemon` organization that I created on Azure. I will help on chat to add the integration to AppDaemon. It took me like 3 minutes to set it up for my fork https://github.com/basnijholt/appdaemon/pull/1 (just as an example).

After the integration is done, each PR should have something that looks like:
image

In order to pass all the flake8 tests I had to fix some bugs
* I added `appdaemon/version.py` to fix #780
* I removed unused imports
* I removed variables that were assigned but never used
* I made some changes to `setup.py` such that AD is only installable with the correct Python (this wasn't the case)
* I've introduced `requirements.txt` to make the installation of *only* the dependencies easier in CI
* Fixing many other small things

Don't be scared of all the changes! The commits tell the story, each one is a separate action. Most of the changes are because of `black` which is done in https://github.com/basnijholt/appdaemon/pull/1/commits/3fb6644087c064ecf9a065a852936163d7d18202 and the changes of the `js` and `css` files are merely changes that remove trailing whitespaces.


acockburn:
Thanks! As discussed, I'll take a look at this and read up on the pipeline etc. then we can take a look when I am ready.

ReneTode:
> @basnijholt temporarily disable other Python versions

do i read that correct. would this disable AD on 3.6 and 3.8?

basnijholt:
@ReneTode, that just disables running the tests on those Python versions, so it has nothing to with AD's compatibility.

As it stands now, there are no real tests that involve running any AppDaemon code. Just `flake8` and `black`, so it makes no sense to run that for different Python versions because it will give the exact same result.

In the future, when there are unit tests (with `pytest`) for example, we can enable the tests for all Python versions again.

Those future tests could for example run some of the example apps for all Python versions that are currently supported.

ReneTode:
ok, thx for clarifying bas.

i see a lot of useless changes, only there to make things really unreadable, but if Andrew wants to go ahead with it, then i guess ill deal with it ;)

acockburn:
Taking a look at this and I like the idea :) I'm thinking of adding it soon after I release 4.0 (today). I have a couple of questions:

1. The black config seems to refer to specific python versions (e.g. 3.7) in pre-commit - we currenlty support 3.6, 3.7 and 3.8 - will this cause an issue?

2. One restriction of PEP8 that I hate and I believe is no longer relevant is the 80 character restriction, for AD I have generally settled on 120. Can this be configured?

I like your requirements.txt  changes so I am adding that manually for the 4.0 release to enable easy future testing with different python versions.

benleb:
line length can be configured, default in black is 88

basnijholt:
Indeed, that is one of the only things that can be configured in `black`.

basnijholt:
@acockburn, I have just merged `dev` and fixed the merge conflicts.

I accidentally missed that you also made a commit here, and force pushed, sorry.

basnijholt:
@acockburn, to answer your questions

> The black config seems to refer to specific python versions (e.g. 3.7) in pre-commit - we currenlty support 3.6, 3.7 and 3.8 - will this cause an issue?

In principle not, but it restricted the user to run black with Python 3.7, I have removed that restriction now though 👍 

> One restriction of PEP8 that I hate and I believe is no longer relevant is the 80 character restriction, for AD I have generally settled on 120. Can this be configured?

I have set the line-length to 120 and reran the filters.

basnijholt:
Great! Now you should connect Azure DevOps to the GitHub repo 😄

acockburn:
Yep - I'll try and catch you online some time to do that :)

PR #828, opened 2020-01-02
added clarification; api: is required for admin

balk77:
added clarification; api: is required for admin interface


ReneTode:
@balk77 its a bug that it is required. so it doesnt need mentioning in the docs, so you can close this.

balk77:
Indeed. See https://github.com/home-assistant/appdaemon/issues/769

PR #831, opened 2020-01-04
Moved asyncio event to be within an async function

Odianosen25:
This became necessary, since the `loop` parameter cannot be passed, as its being depreciated from python 3.10


acockburn:
Can you fix the conflicts please @Odianosen25

PR #865, opened 2020-01-28
Services Lock fix

Odianosen25:
This fixes an issue where a register service, within a service call hangs AD temporarily


acockburn:
Not suer what this is doing ... all it seems to do is move a try statement, it doesn;t seem to have naything to do with locking at all.

PR #871, opened 2020-02-03
Correct logfile to filename

eifinger:
The docs mention ``logfile`` whereas the code shows that the parameter in the dict should be ``filename``.

I figured that was overlooked =)


acockburn:
Thanks!

PR #900, opened 2020-02-29
Fix leap year bug

chadmccune:
Leap years are divisible by 4, except if evenly divisible by 100, unless its also divisible by 400

Year 3000 isn't a leap year, so when replacing year on Feb 29 of a leap year, it crashes

### DIFF
--- a/appdaemon/scheduler.py
+++ b/appdaemon/scheduler.py
@@ -341,7 +341,7 @@ class Scheduler:
 
     def get_next_entries(self):
 
-        next_exec = datetime.datetime.now(pytz.utc).replace(year=3000)
+        next_exec = datetime.datetime.now(pytz.utc).replace(year=3200)
         for name in self.schedule.keys():
             for entry in self.schedule[name].keys():
                 if self.schedule[name][entry]["timestamp"] < next_exec:


homeassistant:
Hi @chadmccune, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;appdaemon#900).

Thanks!

appdaemon/scheduler.py
```diff
@@ -341,7 +341,7 @@ def is_realtime(self):
 
     def get_next_entries(self):
 
-        next_exec = datetime.datetime.now(pytz.utc).replace(year=3000)
+        next_exec = datetime.datetime.now(pytz.utc).replace(year=3200)
```
Teagan42: 1. I can't find anything in git commits/blame that  explain  why they set the year to 300
2. Setting to 3200 seems like a bandaid fix - but it does work
3. Would prefer to find a better solution - butt still I approve

Teagan42: A better solution to this would be to take the current year and ADD a number divisible by 4 - right now, I feel like on non-leap years thtis will cause issues.

```
next_exec = datetime.datetime.now(pytz.utc)
year_offset = next_exec.year + 3200
next_exec = next_exec + year_offset
```

This would make sure leap or not, it gets the right day.

zanfur: Is there a compelling reason not to use datetime.max?  This is the clearly the intended behavior, from the surrounding code.

acockburn:
I ultimately fixed this a slightly different way using `datetime.MAXYEAR`  - thanks for this PR though , I won't merge it but I gave you credit for the emergency fix :)

PR #901, opened 2020-02-29
Plugin event registration

Odianosen25:
This PR allows for plugins to register for events, which can be helpful for plugins willing to make use of such a service. 

This uses the same semantics as the events subscription in ADStream. This way it allows for plugins to use wildcards for subscriptions, making it more flexible compared to having to specify each namespace.


acockburn:
Closing pending fixes in listen_event as discussed

PR #903, opened 2020-03-02
Plugin cleanup

Odianosen25:
Added plugin cleanup, when the plugin is being stopped


appdaemon/plugin_management.py
```diff
@@ -135,6 +132,9 @@ def stop(self):
         self.stopping = True
         for plugin in self.plugin_objs:
             self.plugin_objs[plugin]["object"].stop()
+            name = self.plugin_objs[plugin]["name"]
+            self.AD.loop.create_task(self.AD.callbacks.clear_callbacks(name))
```
acockburn: Why not just await this?

Odianosen25: That is because the caller is done non-async I think. `__main__.py` calls `self.AD.stop()` within thread which in turn calls plugin stop. So thought I should do it this way

acockburn: It's async, yes, but we are shutting down and want to do everything in sequence so please use await here.

Odianosen25: I used `await` and AD doesn't start as it says
```
Traceback (most recent call last):
  File "/home/pi/.local/lib/python3.7/site-packages/appdaemon/__main__.py", line 117, in run
    self.AD = ad.AppDaemon(self.logging, loop, **appdaemon)
  File "/home/pi/.local/lib/python3.7/site-packages/appdaemon/appdaemon.py", line 17, in __init__
    import appdaemon.plugin_management as plugins
  File "/home/pi/.local/lib/python3.7/site-packages/appdaemon/plugin_management.py", line 136
    await self.AD.callbacks.clear_callbacks(name)
    ^
SyntaxError: 'await' outside async function
```
Or should we leave it to the plugins's responsibility to clear itself?

acockburn: OK, let me take a look

Issue #875, opened 2020-02-05
Option To Fetch Latitude / Longitude / Elevation and TimeZone From HA or Environment Variables


danieledwardgeorgehitchcock:
Hi,

I can see in my log files that AppDaemon is now looking for some new parameters (Latitude, Longitude, Elevation and TimeZone) in its configuration file.

It would be beneficial if there was an option to pull these values directly from HA via the following API endpoint:

[Home Assistant Get Config](https://developers.home-assistant.io/docs/en/external_api_rest.html#get-apiconfig)

This will allow for AppDaemon and HomeAssistant to stay in-sync.

Further to this, it would also be beneficial if the parameters could be set / overridden via Environment Variables so that these can be set in Docker Containers, etc.

Apologies if this functionality already exists - I couldn't find it in the documentation..

Odianosen25:
Hello @danieledwardgeorgehitchcock,

The reason it wouldn’t be possible to get these values from HA, is that now AD’s internal schedulers are started before the plying, and dependent  on these values. So if we have the option to get it from HA as was the case before, it just wouldn’t be feasible at the moment.

We can look into the environment variables, but what level do you want to be able to set? The config in appdaemon.yaml?

Regards

danieledwardgeorgehitchcock:
Thanks for the quick reply!

Yes I see the Environment Variables being applied at the config level in appdaemon.yaml or maybe being able to reference them in a similar way to HomeAssistant:

[Home Assistant Environment Variables](https://www.home-assistant.io/docs/configuration/yaml/#environmental-variables)

Odianosen25:
Ok no worries, we will see what we can do, but at this time honestly no promises. Will discuss it with the other devs especially @acockburn, and see what we can come up with.

Regards

Odianosen25:
@danieledwardgeorgehitchcock,

This is now in dev, please check it out and let us know if further stuffs needed

PR #911, opened 2020-03-04
Environ variables

Odianosen25:
This adds the ability to make use of environment variables in AD configuration, instead of secrets. This also fixes the issue #875


danieledwardgeorgehitchcock:
@Odianosen25 Just tried the new !env_var in the Docker dev image - it doesn't seem to work.

Following error message:

```
ERROR could not determine a constructor for the tag '!env_var'
ERROR Error loading configuration
ERROR parser says
ERROR   in "<unicode string>", line 2, column 13:
      latitude: !env_var LATITUDE
                ^
```

danieledwardgeorgehitchcock:
Pulled latest dev image today and all is working!

Issue #910, opened 2020-03-04
get_tz_offset is always returning zero


sillyfrog:
I'm new to AppDaemon, and figuring out time stuff - including timezones - and am finding that `get_tz_offset()` is always returning zero.

Running `self.datetime(True)` gives me a (correct) timezone aware datetime, eg: `2020-03-04 14:04:26.233414+10:00`

For the moment, I'm using the following line to get the number of seconds the timezone is offset:
```
self.AD.tz.utcoffset(self.datetime()).total_seconds()
```

I'm wondering if this wouldn't be a better implementation of the `get_tz_offset` method. (I also noticed that the current implementation does not support 1/2 hour offsets based on the comments, this would also address that).

If you think it makes sense, I'm happy to do a PR, but not sure if there are other implications that I'm missing.

Cheers.

Odianosen25:
Hello @sillyfrog,

Do create the PR please and it will be reviewed. I am sure @acockburn will really be willing to look into it. 

Thanks

acockburn:
What timezone are you in? this API call works fine for me  ...

sillyfrog:
I tried a few. I'm in Australia/Brisbane, but also tried Australia/Sydney and America/Los Angles (this is off the top of my head, I did verify them at the time, and the code mentioned above worked). All by tests were done in the latest docker container.

PR #912, opened 2020-03-05
Correctly return the timezone offset in get_tz_offset

sillyfrog:
As per #910, the current implementation of `get_tz_offset()` is always returning zero. This change addresses that.

That said, after looking at this a bit, I'm thinking that `get_tz_offset()` should probably return seconds rather than minutes (keen for feedback), or a `timedelta` object.

I also noticed that`parse_utc_string()` docstring does not match what it does (I have fixed that). However I think it should probably also go as we have `parse_datetime`, and returning a timestamp which is typically in UTC already feels a bit weird (but maybe I'm misunderstanding how it could be used). I didn't see any uses of it on the internet, so image it's not a big one. It could also raise a `DepreciationWarning`, with a view to remove it down the track.

Keen for any thoughts and happy to integrate them as well.

PS: I also noticed that the code is not formatted with `black` (like HA is), I'm also happy to look at doing that (in another PR), and putting in a config for vscode to do this automatically (also copied from HA).  Cheers.

### DIFF
--- a/appdaemon/adapi.py
+++ b/appdaemon/adapi.py
@@ -1778,22 +1778,14 @@ class ADAPI:
             utc_string (str): A string that contains a date and time to convert.
 
         Returns:
-            An UTC object that is equivalent to the date and time contained in `utc_string`.
+            An POSIX timestamp that is equivalent to the date and time contained in `utc_string`.
 
         """
         return datetime.datetime(*map(int, re.split(r"[^\d]", utc_string)[:-1])).timestamp() + self.get_tz_offset() * 60
 
-    @staticmethod
-    def get_tz_offset():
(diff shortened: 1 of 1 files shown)


homeassistant:
Hi @sillyfrog, 

It seems you haven't yet signed a CLA. Please do so [here](https://home-assistant.io/developers/cla_sign_start/?pr=home-assistant&#x2F;appdaemon#912).

Thanks!

acockburn:
No harm in leaving parse UTC stream as it is - it might break stuff for people. Thanks for the fix!

acockburn:
I didn;t have an issue with the old implementation but if this works for you and the old one didn;t lets go with it ...

PR #943, opened 2020-03-23
Add Sentry support

dcramer:
Sentry can be configured in appdaemon.yaml:

```
sentry:
  dsn: https://key@sentry.example.com/1
```

### DIFF
--- a/appdaemon/__main__.py
+++ b/appdaemon/__main__.py
@@ -151,6 +151,31 @@ class ADMain:
 
             self.logger.info("AppDaemon Exited")
 
+    def setup_sentry(self, config):
+        """Configure Sentry integration.
+
+        Config supports the following attributes:
+
+        - dsn
+        - environment
+        - release (defaults to appdaemon-{version})
+        """
+        import sentry_sdk
+        from sentry_sdk.integrations.logging import LoggingIntegration
+
+        # https://docs.sentry.io/platforms/python/logging/
+        sentry_logging = LoggingIntegration(
+            level=logging.INFO,  # Capture info and above as breadcrumbs
+            event_level=logging.ERROR,  # Send errors as events
+        )
+
+        sentry_sdk.init(
+            dsn=config.get("dsn"),
+            environment=config.get("environment"),
+            integrations=[sentry_logging],
+            release=config.get("release") or f"appdaemon-{utils.__version__}",
+        )
+
     # noinspection PyBroadException
     def main(self):  # noqa: C901
         """Initial AppDaemon entry point.
@@ -389,6 +414,12 @@ class ADMain:
 
         if exit is True:
             sys.exit(1)
+
+        if "sentry" in config:
+            try:
+                self.setup_sentry(config["sentry"])
+            except Exception:
+                self.logger.exception("Unable to setup Sentry")
 
         utils.check_path("config_file", self.logger, config_file_yaml, pathtype="file")
 


dcramer:
I quickly whipped this up to get feedback. This would aid in the same way that it has been helpful to have in hass core.

If you're not familiar with Sentry its an open source-ish application monitoring platform (that I originally built). https://github.com/getsentry/sentry

dcramer:
Aside if this makes sense, I can add docs (pointers to where they make sense would be helpful)

dcramer:
It was also pointed out I could easily do this in my own module - if thats what folks prefer I'm happy to take that recommended approach and we can close this.

ReneTode:
> It was also pointed out I could easily do this in my own module 

thats more logical because then everyone can decide for himself to use it or not.
and at least another server should be configured in the same http section, just a different port

dcramer:
If a plugin makes sense I can go down that road. Also happy to just contribute something to docs that shows how to do this if that'd be helpful?

Odianosen25:
@dcramer,

Well it’s actually @acockburn’s decision to make and not mine. So will reopen it, and wait for his comment on this.

But if I understand what it does, all errors in AD will be caught by this integration and forwarded to a server? As depending on what it does, maybe will make sense to have in core if Andrew thinks it’s needed.

Also I don’t know if you know this, but AD has the ability to stream its log data over stream. So is this in anyway very different.

Apologies if I sound ignorant, as I couldn’t really understand what it does outside forwarding error data to an endpoint for monitoring.

Regards

dcramer:
Yep that’s all it does. I use Sentry to monitor my hass install and it’s quite valuable since most errors are actually my config or integrations. It’s quite a bit different than logging but definitely Sentry specific.

Odianosen25:
@dcramer,

Thanks for the clarifications. If that's the case then I think a plugin is best. AD sends all its log (except `DEBUG` for now) as events. Now we recently allowed for plugins to listen to such events as need be. There is an AD to AD plugin being worked on, which will allow for example logs from one AD to be sent to another in realtime; I believe Sentry based on what you just said works similarly.

So what I would advise (pending what Andrew says) is that make it a plugin, then the plugin logs into the Sentry server. All the plugin will do is listen to the `__AD_LOG_EVENT` and when it gets the data format it in a way Sentry like and forward it as needed.

That's my take on it anyway, but nice idea.

Regards

dcramer:
The one downside here - just so everyone has context - is that Sentry needs more than log messages. It has a logging integration (to capture things like logging.error), but ultimately it needs access to the process when the errors are generated. Here's an example of what Sentry can do (with a bunch of info redacted due to public share link) with exception objects:

https://sentry.io/share/issue/b76f7bdc7f9d4985979c2c37f03c56e4/

dlashua:
@dcramer hey! Sentry looks quite nice! I will have to give it a spin.

That being said, if this can be implemented as a plugin/module it should be. Have a look at the hass plugin as an example ( https://github.com/home-assistant/appdaemon/tree/dev/appdaemon/plugins/hass ) to see if you're able to provide this functionality that way.

If so, please submit a PR with Sentry implemented as a plugin. And in a future release there should be some documentation available on how to create and install a plugin that is not distributed with AD which should allow you to provide this functionality to other users even if it isn't included in the core.

acockburn:
Closing due to inactivity

PR #975, opened 2020-05-13
Select option patch

Odianosen25:
Allow for when using AD generated input select which are just entities, so it sets the right thing

### DIFF
--- a/appdaemon/widgets/baseselect/baseselect.js
+++ b/appdaemon/widgets/baseselect/baseselect.js
@@ -94,7 +94,7 @@ function baseselect(widget_id, url, skin, parameters)
                 }
 
                 else args["option"] = self.state;
-                
+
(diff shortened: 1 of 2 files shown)


ReneTode:
this change is only needed because you create your input_select in AD not the same way as in HA.
logical would be to create the entity the same way or define how an input select in AD should be created.

this way, this change is for your personal use only.

Odianosen25:
Well true. I could possibly just have an app in AD that uses the same thing and translates it in AD. Thinking of it, that’s probably the right way to go, didn’t really reason it before, thanks for the knock.

Will close this and create an app that fixes this locally for me. On another hand for this to work properly, as we have different namespaces will entail having this service call in all namespaces; shouldn’t be a biggy.

Thanks

PR #976, opened 2020-05-14
Dash input boolean patch

Odianosen25:
Make use of the `input_boolean` service call and not `homeassistant`

### DIFF
--- a/appdaemon/widgets/input_boolean.yaml
+++ b/appdaemon/widgets/input_boolean.yaml
@@ -4,10 +4,10 @@ state_active: "on"
 state_inactive: "off"
 enable: 1
 post_service_active:
-    service: homeassistant/turn_on
+    service: input_boolean/turn_on
     entity_id: "{{entity}}"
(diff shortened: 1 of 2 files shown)


ReneTode:
why?

Odianosen25:
@ReneTode ,

So it can use the proper domain service calls. This also allows it to be easily used within AD<  when using AD generated input_boolean entities. This affects nothing from a user perspective. But if wanting to use the entity outside HA, without it having the right domain service call, it becomes a nightmare

ReneTode:
more logical then creating all those services in AD is creating a general one.

dont create:
input_boolean/turn_on
switch/turn_on
etc.

but create: 
appdaemon/turn_on

or else you endup recreating every HA service.

ReneTode:
besides that you can also replace it yourself
just copy the yaml file to the custom_widgets dir and change your service to whatever you like.

so no nightmare at all. and also no limitation to HA.

PR #986, opened 2020-05-31
Fix documentation unit for sleep delay parameter

matejdro:
Documentation states that unit is int, but user can also pass in fractions to sleep for less than a second.


matejdro:
What do I need to do to push this PR further? Any changes / tests I should add?

Issue #1003, opened 2020-07-11
Compatibility with YAML anchors


xaviml:
## Problem

I am opening this issue for a **feature request proposal**. The PyYAML version that AppDaemon uses allows YAML anchors, which is great to reduce code duplication between application configurations. Let's say we have these 2 application configurations:

```yaml
example1:
  module: my_module
  class: my_class
  attr1: common_attr
  attr2: common_attr
  attr3: common_attr
  attr4:  different_value1

example2:
  module: my_module
  class: my_class
  attr1: common_attr
  attr2: common_attr
  attr3: common_attr
  attr4:  different_value2
```

Now with AppDaemon you can reduce this to the following:

```yaml
common_config: &common_config
  module: my_module
  class: my_class
  attr1: common_attr
  attr2: common_attr
  attr3: common_attr

example1:
  <<: *common_config
  attr4: different_value1

example2:
  <<: *common_config
  attr4: different_value2
```

This works perfectly now, however, PyYAML does not remove the `common_config` from the parsed YAML and AppDaemon tries to run the app `common_config`. What I have seen in other projects like [GitLab CICD](https://docs.gitlab.com/ee/ci/yaml/#hide-jobs), for example, is that adding a dot (.) to the beginning of the config ignores that config.

## Feature request

So the feature request would be to ignore configurations starting with a dot, so the following configuration would work as expected:

```yaml
.common_config: &common_config
  module: my_module
  class: my_class
  attr1: common_attr
  attr2: common_attr
  attr3: common_attr

example1:
  <<: *common_config
  attr4: different_value1

example2:
  <<: *common_config
  attr4: different_value2
```

## Addition information

What I am proposing is inserting another if condition in [here](https://github.com/AppDaemon/appdaemon/blob/cd0fa4cd83480480287b6970b9c47e07330cf8cd/appdaemon/app_management.py#L361) like:

```python
and not app.startswith(".")
```

Let me know what you guys think and thank you for this awesome project! :)

Odianosen25:
@xaviml this is a very great suggestion in my opinion, can you just make a PR for it? It looks relatively straightforward as you said.

Regards

xaviml:
This issue can be closed since it was released in AppDaemon v4.0.6.

PR #1004, opened 2020-07-11
feat(app_mgmt): ignore apps starting with a dot

xaviml:
This PR solves #1003. This ignores any app configuration starting with a dot, this has two use cases:
- If a user wants to ignore an application, they can just add a dot at the beginning instead of commenting each line of the YAML config
- Allows ignoring base code configuration when using YAML anchors feature. See more in detail in #1003.

### DIFF
--- a/appdaemon/app_management.py
+++ b/appdaemon/app_management.py
@@ -355,6 +355,13 @@ class AppManagement:
                                             # We don't care what it looks like just pass it through
                                             #
                                             valid_apps[app] = config[app]
+                                        elif app.startswith("."):
+                                            #
+                                            # We ignore any configuration starting with a dot.
+                                            # This could be useful to not comment everything
+                                            # and ignore common code when using YAML anchors.
+                                            #
+                                            pass
                                         elif (
                                             isinstance(config[app], dict)
                                             and "class" in config[app]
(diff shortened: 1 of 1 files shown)


Odianosen25:
Hello @xaviml,

Apologies that has taken a while, but now ready to start doing some merging. Can you kindly edit this, such that as long as there is a dot in the app name, the app will not be loaded? And not just limited to being started with a “.”.

kind regards

xaviml:
Hi @Odianosen25,

Change is made. Just for curiosity, why do you prefer for the app to contain the dot rather than starting with a dot? Won't that break more use cases?

Regards,

Odianosen25:
Hello @xaviml,

normally an app name should not have a dot in it, as the app is converted to an entity in the admin namespace. If the name has a dot, it could break AD internally, and if it’s not now, then a few other changes coming will.

Will work on some extra documentation for this. Hope that explains it?

kind regards

Odianosen25:
What kind of use cases you thinking it will prevent?

xaviml:
Hi @Odianosen25,

I understand now. I was not aware of placing a dot in the app config would break the configuration anyway. My thoughts were more on the line that if anyone is using dots for their app name it was less likely to be at the beginning than just any other places. I was trying to understand what the impact was for users and if you were okay with it.

In my opinion, I am fine either way, I don't think anyone is using dots in their config app names anyway.

Thank you for your explanation :)

Regards,

ReneTode:
>  I don't think anyone is using dots in their config app names anyway

if they try they break AD and come to discord with questions ;) 
using a dot in the appname wasnt possible for a long time. it was just not noticed anywhere ;)

PR #1008, opened 2020-07-18
Www folder

Odianosen25:
This moves the `web` folder to using `www`


ReneTode:
why create a breaking change, that wont add any extra value?

Odianosen25:
Andrew requested it

PR #1016, opened 2020-08-03
WIP: AD Plugin

Odianosen25:
This adds a plugin which allows for two or more AD plugins to be connected using different topologies. It supports the following
- Point-to-point
- Daisy chain
- Star
- Ring
- Mesh etc

Of course depending on the topology selected, the setup will be different.

A sample config is seen below, pending when the docs are available:

```
DEV1:
      type: ad
      ad_url: http://192.168.xxx.xxx:5050 # remote AD's url
      api_key: !secret appdaemon_password # remote AD's password
      ca_certs: ssl.crt
      ca_cert_path: /etc/ssl/certs
      ssl_certificate: ssl.crt
      ssl_key: ssl.key
      namespace: dev1
      client_name: dev1 # client name to be used, when registering on the remote one
    
      remote_namespaces:
        dev1_admin: admin # map remote admin's namespace as "dev1_admin" namespace locally
        dev1_default: default # map remote default's namespace as "dev1_default" namespace locally
        hass: hass # as there is no local "hass" namespace,  map remote hass's namespace as "hass" locally. Useful when the remote AD is the only one connected to HA, and want all other ADs get data from it

      subscriptions: # subscribe to state changes on the remote one locally
        state:
          - namespace: admin # need the remote admin's namespace
            entity_id: "*" # want all entities in the namespace

          - namespace: default # need the remote default's namespace
            entity_id: "*"

         - namespace: hass # need the remote hass's namespace
            entity_id: "*"
        
        event: # subscribe to events on the remote one locally
          - namespace: default
            event: "*"

         - namespace: hass
            event: "*"
``` 

The checks that fail is due to the `__init__.py`, and that will need fixing when merging.


benleb:
oh that would also enable some kind of failover (maybe not directly but via an app for example), right? 👌

ReneTode:
that can be done without an AD-AD plugin.

Odianosen25:
@benleb,

Yes that’s true, but as @ReneTode said, you don’t exactly need the plugin for that. The main reasons I built it, was:
- I when I got two AD instances running different programs due to complexity and load, I need them To work as a single instance. 

- I got AD instance in Azure CI, so the local one connects to the one in the cloud, without doing port forwarding and send instructions 

It allows for other more complex stuffs, which I am working on, but over time will possibly release them. Andrew likes this plugin, and opted to add it to the core, so I am making a PR. Been on it for over a year now

acockburn:
Closing as it has conflicts. Will re-open when you have time to work on this.

PR #1018, opened 2020-08-07
Widget: clickable icon

JakubMacoun:
This widget combines existing widgets: _icon_ and _script_. It shows an icon same as the _icon_ widget and adds the ability to run a HA script on click which should perform desired actions possibly driven by a state of another entity (usually the one monitored by the icon).
Another feature added to this widget is a possibility of adding update delay - when set, it delays state update by the specified amount of seconds (useful when it takes some time to update secondary states by HA - in that case, this can prevent blinking of the icon).

Usage example:
```
ac_control:
    widget_type: clickable_icon
    entity: sensor.ac_db_control_option
    script_entity: script.toggle_ac
    icons:
        "turn_on_temporarily":
            icon: mdi-air-conditioner
        "turn_on":
            icon: mdi-air-conditioner
        "turn_off":
            icon: mdi-air-conditioner
            style: "color: aquamarine;"
    state_text: 1
    state_map:
        "turn_on_temporarily": Turn on AC temporarily
        "turn_on": Turn on AC
        "turn_off": Turn off AC
```

### DIFF
--- a/appdaemon/widgets/baseclickableicon/baseclickableicon.css
+++ b/appdaemon/widgets/baseclickableicon/baseclickableicon.css
@@ -1,0 +1,32 @@
+.widget-baseclickableicon-{{id}} .title {
+	position: absolute;
+	top: 5px;
+	width: 100%;
+}
+
+.widget-baseclickableicon-{{id}} .title2 {
+	position: absolute;
+	top: 23px;
+	width: 100%;
+}
+
+.widget-baseclickableicon-{{id}} .state_text {
+	position: absolute;
+	bottom: -3px;
+	width: 100%;
+}
+
+.widget-baseclickableicon-{{id}} .icon {
+	position: absolute;
+	top: 43px;
+	width: 100%;
+}
+
+.widget-baseclickableicon-{{id}} .toggle-area {
+	z-index: 10;
+	position: absolute;
+	top: 0;
+	left: 0;
+	width: 100%;
+	height: 100%;
+}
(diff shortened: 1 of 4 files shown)


Odianosen25:
@JakubMacoun,

Thanks for this widget. Just something to consider, as instead of using `script`, why not use AD's inbuilt sequence, which will allow to use both HA's scripts and other custom services that are available within AD.

Regards

ReneTode:
there is no new widget needed for this, because it was already possible.

to have such a thing you can use the existing switch widget and use state map and service settings.

JakubMacoun:
> there is no new widget needed for this, because it was already possible.
> 
> to have such a thing you can use the existing switch widget and use state map and service settings.

Not really, switch allows to use only two states - on and off. Clickable icon can use as many states as needed.

ReneTode:
you are wrong about that.
you can set the parameters state active and state_inactive
so you can use any value you like with a switch widget.

ReneTode:
the only thing is that a switch widget only has 2 states.

so if you would create an acceptable widget that would be in line with others, then it needs the following:

the possibility to set ANY service for ANY state from the sensor.

in that case i would advise to take the icon widget and add the possibility to click it there.
(so no new widget, but an addition to the options from the icon widget)

JakubMacoun:
> the only thing is that a switch widget only has 2 states.
> 
> so if you would create an acceptable widget that would be in line with others, then it needs the following:
> 
> the possibility to set ANY service for ANY state from the sensor.
> 
> in that case i would advise to take the icon widget and add the possibility to click it there.
> (so no new widget, but an addition to the options from the icon widget)

Yep, that sounds like a better solution, therefore I am closing this PR and I will try to work on extending the icon widget. I'll open new PR when it's done.

PR #1022, opened 2020-08-11
Multi-arch Docker images via Github Actions

benleb:
I use this to build a Demo Container with AppDaemon and my [AutoMoLi](https://github.com/benleb/ad-automoli) app (check the gh-action in the repo for a live example). What about using this ultra-simple way to build images also for AppDaemon itself?


benleb:
superseded by #1080 👍

PR #1043, opened 2020-09-22
Add support for installing user-specified system packages with docker run

sdlynx:
Just added something simple. I was encountering some issues where python packages specified in `requirements.txt` were missing system-level dependencies (e.g., `zlib`, `libjpeg`) and wanted to add this in so I could specify custom packages that were required to ensure the `pip3 install` command succeeded.

Let me know if I need to make any tweaks - first time trying to contribute to this repo.

### DIFF
--- a/dockerStart.sh
+++ b/dockerStart.sh
@@ -42,6 +42,9 @@ if [ -n "$DASH_URL" ]; then
   fi
 fi
 
+#install user-specific packages
+apk add --no-cache $(find $CONF -name system_packages.txt | xargs cat | tr '\n' ' ')
+
 #check recursively under CONF for additional python dependencies defined in requirements.txt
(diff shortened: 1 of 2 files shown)


Odianosen25:
Hello @sdlynx,

This is good idea, but a couple of changes I will suggest

- Kindly change the name to `system_packages`. As that sounds more intuitive than `APK_ADD`
- I think this should be in some kind of files, considering the fact its possible to try installing packages in different repository. So maybe putting it in some yaml or something so its list?
- In addition to the above, it makes it neater so the environment variable isn't such a long stuff, and allow for easy editing the file and recreate the container when adding stuffs

Regards

sdlynx:
Thanks for the feedback! Made the suggested changes (and squashed). Let me know if you have any other thoughts.

sdlynx:
Bumping this so it gets merged!

PR #1065, opened 2020-10-31
Update CONFIGURE.rst

chbndrhnns:
The inital configuration provided in the docs should work out of the box. The additions I made allow for that.

### DIFF
--- a/docs/CONFIGURE.rst
+++ b/docs/CONFIGURE.rst
@@ -15,6 +15,10 @@ Your initial ``appdaemon.yaml`` file should look something like this if you are 
 .. code:: yaml
 
      appdaemon:
+       time_zone: CET
+       latitude: 51.725
+       longitude: 14.3434
+       elevation: 0
        plugins:
          HASS:
            type: hass


ReneTode:
adding the wrong settings to the AD config is the wrong way.
there should be a check in the startup from AD that checks correctly if the settings are added or not, and if not there should be a warning in the logs.

AD with default wrong settings is just as useless as AD that doesnt start at all.

chbndrhnns:
Would it work if there was a note about replacing the example values with the actual ones?

ReneTode:
i think AD just should stop if no values are detected.
only at this point it gives errors where it should give a decent text in the logs.

putting in example values will only make that people who dont look at the logs afterwards start complaining that AD is running in the wrong time.

in my eyes AD should never start if the needed information isnt given.

chbndrhnns:
My take on that matter is that I'd like to go through the docs once when I am setting up a new system. This is obviously not enough in the case of AD as certain required configuration parameters are missing from docs.

ReneTode:
the required configuration is all in the docs.
but the docs good be made better at some places, if you find things missing, or wrong then please make a PR (as you did with the .cfg what has changed to yaml)

the appdaemon.yaml example should not contain random settings, but empty settings that have remarks.
and then in the startup from the python AD should check if those settings are filled, and when not then it should log a warning and stop AD.

at least thats my idea how it should work.

Odianosen25:
I am not sure what the conversation here is for, but I was about to merge this, so is there something I am missing?

@chbndrhnns, thanks for this. The data you put in there as default, where does it lead to?

Regards

chbndrhnns:
I believe we have a different understanding on the documentation approach. Mine is that, reading the documentation, I would like to be able to copy some sample snippet and make it work. This is why I added some example values to the CONFIGURE.rst.

René, however, believes that no wrong values should be in the docs (like a time zone that is not the user's timezone).

ReneTode:
my bad. i didnt really read this PR correctly.
i thought it was about the code, and not the docs.

in the docs examples can have anything, as far as i see it.

in the code it should be empty after install. but some also want AD to start with some example values, instead of AD just stopping because there are no correct values given.

PR #1090, opened 2020-12-08
Docs - Better Callback Documentation

rr326:
Created a Callback Signature section of AD AD_API_REFERENCE

In learning to use AD I found myself constantly jumping back and forth
to the callback signatures for events / states / scheduler.  

The current process is quite cumbersome:

1. Goto AD API 
2. Open Reference
3. Go to your main area (state, scheduler, event)
4. Scroll to the listen event
5. Click to where the callback is defined in the guide.

This would have made my life much easier!

BTW - Perhaps a better way to solve this would be to have the full
documentation here in the API page. Maybe you could write the
single source of truth in the code and reference it both 
in the API doc and the Guide. But I didn't want to do a bunch
of work only to have it be wrong.

### DIFF
--- a/docs/AD_API_REFERENCE.rst
+++ b/docs/AD_API_REFERENCE.rst
@@ -264,6 +264,37 @@ WebRoute
 .. autofunction:: appdaemon.adapi.ADAPI.register_route
 .. autofunction:: appdaemon.adapi.ADAPI.unregister_route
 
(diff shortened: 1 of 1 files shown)


Odianosen25:
Please open this against dev branch and not master

rr326:
Replaced with #1097

PR #1095, opened 2020-12-12
The documentation of listen_event() pointed to the state callback sig…

rr326:
The documentation of listen_event() pointed to the state callback signature. Fixed.


Odianosen25:
Thanks @rr326

PR #1097, opened 2020-12-12
Created a Callback Signature section of AD AD_API_REFERENCE

rr326:
In learning to use AD I found myself constantly jumping back and forth
to the callback signatures for events / states / scheduler.

The current process is quite cumbersome:

1. Goto AD API
2. Open Reference
3. Go to your main area (state, scheduler, event)
4. Scroll to the listen event
5. Click to where the callback is defined in the guide.

This would have made my life much easier!

BTW - Perhaps a better way to solve this would be to have the full
documentation here in the API page. Maybe you could write the
single source of truth in the code and reference it both
in the API doc and the Guide. But I didn't want to do a bunch
of work only to have it be wrong.


Odianosen25:
Hello @rr326,

pls can you add the listen log callback to the callbacks section and also rebase and clear checks so I can merge this?

regards

rr326:
@Odianosen25 Working on it. Can you point me to the place where it shows how to rebase properly? (I've got it rebased, but I  think I need to do a push -f, and I don't want to mess it up.)  I looked around but can't seem to find those instructions.

rr326:
@Odianosen25 Also, I can't find the listen_log() callback signature anywhere in the docs. I'll create one from looking at the code, but if it is defined somewhere and I'm just missing it, please point me to it so I know I get it right.

rr326:
@Odianosen25 Actually, I'm not sure I can figure out the callback signature. I think it is the following:

```python
def my_callback(self, log_data, kwargs):
    <do some useful work here>
```

Odianosen25:
Hello @rr326,

I only just noticed we had no docs on log callback 🙄. Well see if it’s something that can be updated later. Maybe after then; we can update yours.

Regards

rr326:
I'm pretty certain that the above is the right signature. I might as well put that up with no "Full documentation <link>".  Happy to do that if you want. But then I'd need those pull request instructions.

Odianosen25:
Actually this is it as I have it in one of my apps

```python
def log_callback(self, name, ts, level, log_type, message, kwargs):
        data = json.dumps(
            {
                "name": name,
                "ts": str(ts),
                "level": level,
                "log_type": log_type,
                "message": message,
            }
        )
```

Regards

rr326:
@Odianosen25 I added the signature (below) on a rebased branch and then did `push -f`, which may have been a mistake? It's now not updating THIS PR but seems to be asking for another PR. (Which, btw, is also failing due to a docker build problem.) Let me know how to proceed.

Here is the sig I wrote:

```rst
**Log Callback**

.. code:: python

      def my_callback(self, app_name, timestamp, log_level, log_name, log_message, kwargs):
        <do some useful work here>
        
Full documentation - *TBD*
```

Odianosen25:
> I added the signature (below) on a rebased branch and then did push -f, which may have been a mistake?

Yeah there was no need to do that, but not sure what the consequences are. But shouldn't e too bad, can always reset it.

> It's now not updating THIS PR but seems to be asking for another PR

You may need to ensure you are working on the right branch locally, and do a `git pull` before continuing. So something like the below

```
git checkout dev_better_callbacks
git pull
```

After the above, your remote and local should be same. If `git pull` is failing, then depending on if you have done a lot of work on the files presently, you could

```
git reset --hard # this will reset your local to the last merged
git pull # try to make the remote and local same
```
Now depending on which branch you made the last changes, you could `rebase` again on that branch if and only if its up to date on `dev`

Then you can work from there. I honestly haven't been in a situation of the side effects of using `pus -f`, I have tried it and it didn't work so never got to fixing whatever issues that may have arose from there.

Just let me know how it goes.

Regards

rr326:
@Odianosen25 My bad - I was checking in the wrong branch.  (My repo is a mess. Once you've processed all my PRs I'll clean it all up and be better going forward.) 

The build is failing but that is some azure thing - I don't think that has anything to do with me.

acockburn:
No activity - closing

Issue #1064, opened 2020-10-31
Appdaemon is not compatible with Python 3.9


chbndrhnns:
Running it in a 3.9 venv gives me this error:

```
Traceback (most recent call last):
  File "/Users/jo/.virtualenvs/appdaemon-MLpxRBnC/bin/appdaemon", line 5, in <module>
    from appdaemon.__main__ import main
  File "/Users/jo/.virtualenvs/appdaemon-MLpxRBnC/lib/python3.9/site-packages/appdaemon/__main__.py", line 23, in <module>
    import appdaemon.http as adhttp
  File "/Users/jo/.virtualenvs/appdaemon-MLpxRBnC/lib/python3.9/site-packages/appdaemon/http.py", line 9, in <module>
    import feedparser
  File "/Users/jo/.virtualenvs/appdaemon-MLpxRBnC/lib/python3.9/site-packages/feedparser.py", line 93, in <module>
    _base64decode = getattr(base64, 'decodebytes', base64.decodestring)
AttributeError: module 'base64' has no attribute 'decodestring'
```

benniju:
I have exactly the same Problem running with pyenv on Linux with a Raspberry Pi:

```
Traceback (most recent call last):
  File "/home/homeassistant/.pyenv/versions/appdaemon/bin/appdaemon", line 5, in <module>
    from appdaemon.__main__ import main
  File "/home/homeassistant/.pyenv/versions/3.9.0/envs/appdaemon/lib/python3.9/site-packages/appdaemon/__main__.py", line 23, in <module>
    import appdaemon.http as adhttp
  File "/home/homeassistant/.pyenv/versions/3.9.0/envs/appdaemon/lib/python3.9/site-packages/appdaemon/http.py", line 9, in <module>
    import feedparser
  File "/home/homeassistant/.pyenv/versions/3.9.0/envs/appdaemon/lib/python3.9/site-packages/feedparser.py", line 93, in <module>
    _base64decode = getattr(base64, 'decodebytes', base64.decodestring)
AttributeError: module 'base64' has no attribute 'decodestring'
```
I guess I will have to work with Python 3.8, which is annoying because Home-Assistant works on 3.9 (of course I have a seperate venv)

ReneTode:
its always difficult to keep up with other programs.
HA has a ton of (partly paid) developers, AD only has a few.
HA decided that users need to upgrade to python 3.8, while AD still is running on python 3.6 (which is very fine for synology users)

[...]

PR #1103, opened 2020-12-27
python3.9 compatibility, #1064

daringer:
* fixes the deprecated `all_tasks` call
* not save for 3.10 as of now (passing `loop` will be deprecated)
* not passing `loop` leads to an error as the internally called `get_loop` fails for some reason
* **warning** this likely will break all python <=3.9 versions

### DIFF
--- a/appdaemon/__main__.py
+++ b/appdaemon/__main__.py
@@ -130,7 +130,11 @@ class ADMain:
 
             self.logger.debug("Start Main Loop")
 
-            pending = asyncio.Task.all_tasks()
+            try:
+                pending = asyncio.Task.all_tasks()
+            except AttributeError:
+                pending = asyncio.all_tasks(loop)
+
             loop.run_until_complete(asyncio.gather(*pending))
 
             #


ReneTode:
> warning this likely will break all python <=3.9 versions

in that case this shouldnt be merged!!

daringer:
yes, please do not merge for now. It generally was not clear (from the issue) for me, if such a change is wanted upstream.

But I can easily make this a proper PR to be merged to upstream, should I?
Therefore I would just insert a python version-check, anything else despite this to make it mergable?

ReneTode:
if you submit a PR that doesnt break older python versions, but it fixes newer python versions, i suspect that noone will object.
but please in that case test and make sure that AD doesnt break on older python versions.

on the other hand it is just to remove a deprecation warning, which i think is harmless.
so it doesnt make AD more compatible with python 3.9, it just prepares it for 3.10 and higher.

daringer:
> so it doesnt make AD more compatible with python 3.9, it just prepares it for 3.10 and higher.
current `dev` is not working with python 3.9 without this PR, so it kind of does both

updated PR, could be merged from my side, please check. 
It's a minimal change and does not explicitly check for version, but simply *tries* both versions.

Happily taking advice and updating, if I missed something

ReneTode:
> current dev is not working with python 3.9 without this PR, so it kind of does both

thats strange, because the warning says its deprecated for 3.10

> It's a minimal change and does not explicitly check for version, but simply tries both versions.

as far as i can see, you changed it so that it only tries the new way when the old way doesnt work.
i can see no objection to that on its own, as long as the devs make a note that it needs to be removed somewhere in the far future (im not a dev anymore, just a guy that is very involved with AD ;) )

> Happily taking advice and updating, if I missed something

i cant think of something you missed, because i cant oversee the scope from such changes.
so i leave that up to the devs.

Odianosen25:
Hello @daringer,

I am sorry I will be closing this PR. I have tested it and as much as it solves the issue of 3.9, but since the loop has to be passed which will be deprecated in 3.10, it really not a sustainable solution.

If you are able to solve the other issues, do kindly reopen it.

Regards

PR #1118, opened 2021-01-09
Use python:3.8 instead of alpine

### LINKED ISSUE #1073: Offer alternative Docker image not based on Alpine (for example Ubuntu)
Using Alpine for Python based applications can be a real pain.

https://pythonspeed.com/articles/alpine-docker-python/

For example, if I use the unmodified AppDaemon image and I add these Python dependencies to `requirements.txt`:

```
numpy==1.19.4
scipy==1.5.4
```

This is *slooooow* since we can't get pre-compiled libraries on Alpine (due to musl libc) and additionally it ultimately fails due to missing dependencies.

In other words, there is no way for me to install these libraries without modifying the Docker image :sigh:

If the image was based on Ubuntu instead, both these packages would install in seconds since they have pre-compiled versions.

DTTerastar:
Fixes #1073

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.8-alpine
+FROM python:3.8
 
 # Environment vars we can configure against
 # But these are optional, so we won't define them now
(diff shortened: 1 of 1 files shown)


Odianosen25:
The logic make sense why to move to Alpine, but will be something we need to discuss it internally.

Thanks for the update.

acockburn:
This is moot since we are upgrading to python 3.9

PR #1131, opened 2021-01-24
rebased Pr/963 just added windows compat

Hellowlol:
So i could test the features.

I tested notify, turn_on, turn_off and toggle. Works fine. All credits to the original author :)

### DIFF
--- a/appdaemon/__main__.py
+++ b/appdaemon/__main__.py
@@ -7,22 +7,30 @@ also creates the loop and kicks everything off
 
 """
 
-import sys
 import argparse
+import asyncio
 import os
 import os.path
-import signal
 import platform
-import yaml
-import asyncio
-import pytz
-import pid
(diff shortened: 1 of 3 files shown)


Hellowlol:
I got `
``2021-01-24 21:03:27.789590 WARNING AppDaemon: Excessive time spent in utility loop: 21551.0ms, 21551.0ms in check_app_updates(), 0.0ms in other when`` when i added some asyncio.sleep() to my manual tests, i don't know if  that a new issue or not.

Odianosen25:
@Hellowlol,

Thanks for this, and will look into it.

As for the warning you gave, its normal if you have a lot of app files and of cause that's dependent on your SSD.

But thanks again for this, and will be merging this to dev.

Also thanks for the fix on uvloop.

Regards

Issue #1146, opened 2021-02-06
External commands / HADASHBOARD not listening for events?


clyra:
Hi,

The docs said that you can navigate to another page using events (https://appdaemon.readthedocs.io/en/latest/DASHBOARD_CREATION.html?highlight=external%20command#external-commands). But going to the homeassistant developer tools->events and firing a hadashboard event doesnt seem to work. I tried with the following:

 ```
   Event type
        hadasboard

    Event Data (Yaml, optional)
      command: navigate
      timeout: 10
      target: Cozinha8-main
      sticky: 0
```

I couldnt find anything about this event on the appdaemon logs.  There's also this post on HA forum describing a similar problem:  https://community.home-assistant.io/t/changing-dashboards-with-scripts-forcing-apdeamon-to-change-skins-remotly/262946

Also, as I understood from the docs this should change the page on all devices dashboards, right? is there a way to limit which device should receive the "navigate" command?  perhaps using namespaces?

ReneTode:
because of changes in HA i have seen several people telling that it doesnt work anymore when you fire the event in HA.
the workaround for that is a simple app in AD that listens to a boolean, automation or script in HA.

an example is somewhere on the HA forum.

there is no way to change 1 single dashboard on 1 device. that is simply not possible because of the way http works.
the only workaround for that that i have found is by using fully kiosk.

clyra:
I did a little digging into this and couldn't find a reference to something like ["event_type"] == "hadasboard" in the code, but I did find that the dash_navigate function fires a event named "__HADASHBOARD_EVENT" into ad event bus. Changing the event type in HA developers tools/events to "__HADASHBOARD_EVENT" doest the trick and the dashboard changes as expected. Also on appdaemon/assets/javascript/dashboard.js there is a line that reads "self.stream.listen_event('*', '__HADASHBOARD_EVENT', self.update_dash); ".   So i guess that somewhere along the development this has changed and it didnt got to the docs.

acockburn:
Thanks for figuring that out - we'll leave this open to remind us to fix the docs.
________________________________

ReneTode:
thx.
yes you are right, the eventname has changed, and at that moment probably the fact that its available for users was forgotten.

thx. for looking into it.

clyra:
[...]

PR #1159, opened 2021-02-13
navigating based on deviceid

clyra:
As discussed here (https://github.com/AppDaemon/appdaemon/issues/1146), this PR adds the creating of a device id and storing it using the js localStorage function. The device id can been seen on the web console. Later this can be used with navigate event to target a specific device by adding "deviceid: <deviceid>". 

The device id can also be show using the custom widget "device id" which can be found here:  https://github.com/clyra/appdaemon-custom-widgets/tree/master/custom_widgets


ReneTode:
i rather see a widget (or other option) that SETS a device id, then a widget that shows an automated device ID.

clyra:
another attempt coming.

PR #1160, opened 2021-02-14
adding deviceid and dashid to the navigate function

clyra:
This PR adds two optional parameters to the navigate function: deviceid which looks for a id stored in a device localstorage cookie (deviceid is auto generated, but can be set using deviceid widget (https://github.com/clyra/appdaemon-custom-widgets); and dashid which matches a substring in dashboard title (if you set dashid to "floor1" all devices on dashboards "floor1 <foo bar>" will navigate.

Small changes were made to dashboard.js and adapi.py


ReneTode:
this would make me happy when the function get_set_deviceID would not just sets an automated ID, but uses a prompt to ask for the ID when its not set.

```
function get_set_deviceID() {
  if(!localStorage[ID_STORAGE_KEY])
  {
      localStorage[ID_STORAGE_KEY] = prompt("Please give a device ID", "");
  }
  return localStorage[ID_STORAGE_KEY];
};

that would make that every device gets a set ID, and there would be no need to lookup the ID anywhere.

and this needs to be documented also in the dashboard section from the docs, to make the PR complete.

clyra:
This function just make sure that there is a value there. You could set a value by using the deviceid widget. Forcing the user to input a value doesnt work for my use cases (TV without easy input device or "magic mirror"). I was wondering if i should submit the deviceid widget as part of the PR or leave it as a custom_widget.

clyra:
docs updated. the deviceid widget inclusion remains to be decided.

ReneTode:
>  Forcing the user to input a value doesnt work for my use cases (TV without easy input device or "magic mirror")

ok i can understand that, but i dont want to be forced to put a widget on my dashboard to set an ID, because thats illogical too.
so in that case there should be an option in appdaemon.yaml that gives people the option to set a device ID manually or automaticly. or even better it should be a popup with a timeout. if after 30 secs no key is given, the popup dissapears and an autokey is generated.

clyra:
Well... doing timeout on a prompt doesnt seen to be easily done on javascript (at least not for this beginner) :-(. A middle ground would be something like this:

   localStorage[ID_STORAGE_KEY] = auto_id;
    var id =  prompt("Please enter a device ID or just reload this window to accept the default value:", auto_id); 
    if (id != null && id != undefined)
    {
        localStorage[ID_STORAGE_KEY] = id;
    }

does this works for you?

ReneTode:
if that would work for you then i got no problem with that.
it will still prompt the first time, so for devices where you can give the ID its what you want, and after the second load the id is what you have set or autogenerated.

indeed a timeout on a prompt isnt easy, so then it would be a more complicated popup. but this solution is simple and elegant enough.

so if @acockburn and @Odianosen25 think it is a good idea then i certainly give this my thumbs up!

clyra:
great! so, let's wait for them.

Odianosen25:
Hello @clyra,

Thanks for this. But can you kindly start by rebasing the PR against the present dev? As not sure why it seems to be against an old version of the dev and maybe that's why it fails.

Regards

clyra:
Sure, but i dont know exactly what I have to do :-(. I had a old fork of appdaemon, and tried to update it to dev, but I guess i messed it. Do you have any pointers, or what i should look for? thx

PR #1163, opened 2021-02-16
Adding deviceid and dashid to external commands

clyra:
This PR adds two optional parameters to the external commands function: deviceid which looks for a id stored in a device localstorage cookie (deviceid can be autogenerated or defined by the user) and dashid which matches a substring in dashboard title (if you set dashid to "floor1" all devices on dashboards "floor1 " will navigate.

I changed the logic a little bit so deviceid and dashid could be used for other commands (not only navigate).

I rebased my branch to match appdaemon:dev>. I also updated the docs, fixing the __HADASHBOARD_EVENT on external commands and a "timeout" repeated in the docs.


Odianosen25:
Hmmm you have merged my PR and yet still failing 🤔. Did you add extra stuff?

Odianosen25:
@ReneTode will you be able to review this and approve so I merge? I personally I am cool with it, but need a second view. I know we been discussing it on Discord, just need extra confirmation. 

Regards

Odianosen25:
Also I am in talks with Andrew, for us to navigate from using `__HADASHBOARD_EVENT`, as its really not user friendly to do it from HA that way, and maybe we could use `AD_DASHBOARD` instead.

He will possibly comment on this when he is available.

Regards

Odianosen25:
@clyra,

Please can you do a couple of things to this PR?

- Leave the `__HADASHBOARD_EVENT` in there as it is in the `dashboard.js` file
- In the other places, kindly replace `__HADASHBOARD_EVENT` with `ad_dashboard` 
- The new event will now be used in the `navigate` api, and please expose this new event in the docs only
- This is more user friendly, and so users should use this `ad_dashboard` instead of the `__HADASHBOARD_EVENT`
- So in essence in `dashboard.js`, both `__HADASHBOARD_EVENT` and `ad_dashboard` will cause the page to navigate
- `__HADASHBOARD_EVENT` is only left in there, for backward compatibility for those already using it.
- 
Hope it makes sense?

Regards

clyra:
Done. __HADASHBOARD changed in dashboard.js, adapi.py and docs.   dashboard.js still listen to __HADASHBOARD_EVENT and will accept it or ad_dashboard.

PR #1170, opened 2021-02-20
Update scheduler.py

wernerhp:
Fix astral import

### DIFF
--- a/appdaemon/scheduler.py
+++ b/appdaemon/scheduler.py
@@ -2,7 +2,6 @@ import traceback
 import datetime
 from datetime import timedelta
 import pytz
-import astral
 import random
 import uuid
 import re
@@ -12,6 +11,7 @@ from collections import OrderedDict
 
 import appdaemon.utils as utils
 from appdaemon.appdaemon import AppDaemon
+from astral.location import Location, LocationInfo
 
 
 class Scheduler:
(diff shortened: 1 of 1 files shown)


Odianosen25:
What’s the major difference between the previous PR and this? And did you run pre-commit so it passes the checks?

regards

Odianosen25:
@wernerhp can you pls correct the checks?

wernerhp:
Sorry, I made the edit on Github directly.  Will update the PR and fix checks.  

I skipped the `flake8` `pre-commit` hook.  For some reason it fails with
```
appdaemon/scheduler.py:19:26: E999 SyntaxError: invalid syntax
```
which is `ad` in 
```
def __init__(self, ad: AppDaemon):
```

wernerhp:
The previous PR imported `astral` and tried to ini `astral.location.Location`, but this doesn't actually work.  Had to change the import and then also imported the `LocationInfo` type for the passed `tuple`.

wernerhp:
Don't really get why the build is failing 🤔

Odianosen25:
@wernerhp can you separate the PR for scheduler fix from the others? So it’s easier to know what is what?

wernerhp:
> @wernerhp can you separate the PR for scheduler fix from the others? So it’s easier to know what is what?

Yeah, sorry accidentally committed my virtualenv directory.  Fixed now.

Odianosen25:
Thanks but the flake pls make another PR so we can handle it separately.

Odianosen25:
Is the “Location” needed after importing “LocationInfo”?

wernerhp:
Yeah, we are still calling `Location`, that's what we want.  `LocationInfo` was just added to define the `tuple` which is passed as an parameter to `Location`'s constructor.  See [`Location.__init__`](https://github.com/sffjunkie/astral/blob/3282d7b437349bc035c8a7dff5c27ec9997a2cdf/src/astral/location.py#L20)

PR #1211, opened 2021-03-10
Enable mouse events

Odianosen25:
This PR allows for sending to AD backend standard events like mouse events from a widget. This allows for registering and making use of the standard jquery callback and get the entire event data, which can then be processed and sent to AD as needed. This way, not being limited to only the data based by dashboard.

### DIFF
--- a/appdaemon/assets/javascript/dashboard.js
+++ b/appdaemon/assets/javascript/dashboard.js
@@ -416,7 +416,7 @@ var WidgetBase = function(widget_id, url, skin, parameters, monitored_entities, 
 
             }
 
-            else 
+            else
             {
                 $(callbacks[i].selector).on(callbacks[i].action, (
                     function (callback, ch, params) {
--- a/appdaemon/assets/javascript/dashboard.js
+++ b/appdaemon/assets/javascript/dashboard.js
@@ -406,13 +406,26 @@ var WidgetBase = function(widget_id, url, skin, parameters, monitored_entities, 
     {
         if ("selector" in callbacks[i])
         {
-            $(callbacks[i].selector).on(callbacks[i].action, (
-                function (callback, ch, params) {
-                    return function () {
-                        callback(ch, params)
-                    };
-                }(callbacks[i].callback, child, callbacks[i].parameters))
-            );
+            if ("event" in callbacks[i] && callbacks[i].event === true)
+            {
+
+                var data = {};
+                data.parameters = callbacks[i].parameters;
+
+                $(callbacks[i].selector).on(callbacks[i].action, data, callbacks[i].callback);
+
+            }
+
+            else 
+            {
+                $(callbacks[i].selector).on(callbacks[i].action, (
+                    function (callback, ch, params) {
+                        return function () {
+                            callback(ch, params)
+                        };
+                    }(callbacks[i].callback, child, callbacks[i].parameters))
+                );
+            }
         }
         else if ("observable" in callbacks[i])
         {
--- a/appdaemon/assets/javascript/dashboard.js
+++ b/appdaemon/assets/javascript/dashboard.js
@@ -406,7 +406,7 @@ var WidgetBase = function(widget_id, url, skin, parameters, monitored_entities, 
     {
         if ("selector" in callbacks[i])
         {
-            if ("event" in callbacks[i] && callbacks[i].event === true)
+            if ("DOMEventData" in callbacks[i] && callbacks[i].DOMEventData === true)
             {
 
                 var data = {};
--- a/docs/HISTORY.rst
+++ b/docs/HISTORY.rst
@@ -6,6 +6,7 @@ Change Log
 
 **Features**
 
+- Added the ability to allow widget to make use of mouse events, which can then be sent to AD backend
 - Added the ability to add custom javascript code to dashboard - contributed by `Rene Tode <https://github.com/ReneTode>`__
 - Added the ability to set `deviceId` on dashboard, via the dashboard URL - contributed by `clyra <https://github.com/clyra>`__
 - Added the ability to navigate to different dashboards on different devices based on the set `deviceId` - contributed by `clyra <https://github.com/clyra>`__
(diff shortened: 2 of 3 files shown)


ReneTode:
why is this change needed, wasnt it already possible to call all html events?

Odianosen25:
It was possible, but one doesn't get the whole event as in the callback, as dasboard forces what data you get. This allows for everything to be passed back as in raw jquery callback. This way, one can do a whole lot more as required. 

Unless I am getting something wrong

acockburn:
Sending front end interactions back to a server is not a really good architectural approach - any local user interactions should be handled by the frontend (e.g. the widget itself, and communication with the server should occur only if the frontend interaction requires some kind of state change or service call. 

I can see value in allowing widgets to register for more types of events including mouse events but the language of the PR seems to imply they will be sent back to AD which I don't agree is necessary - however the code doesn't seem to do this. Can you please clarify?

Odianosen25:
Hello @acockburn,

It wouldn't by default be sent to AD, my bad as I didn't clarify. It allows the widget to register to get the event object jquery sends when an event happens, then within the widget it can then be decided what should be sent to the backend. Its not intended to send raw data from user interactions as it may to the backend.

My use case was I needed mouse events sent back as events to AD. But using the default callback, there was no way I could extract mouse co-ordinates (please correct me if wrong) and for that reason I needed the event object supplied by jquery. So within the widget, I then process the data needed, and using a `call_service` fire the event as need be back to AD.

This is how I do it in my custom widget. If there is a way to do it without this PR, will be happy to use that

```javascript
var callbacks = [
        {"selector": "img", "action": actions, "event" : true, "callback": OnEvent},
    ];


function OnEvent(event)
    {
        console.log(event);
        
        var x_pos = event.offsetX;
        var y_pos = event.offsetY;

        var mouse_event = event.type;

        // now decide if to process the event depending on if its a mousemove
        // and if the mouse if over the minimum threshold

        if (mouse_event == "mousemove" && 
            (((x_pos + y_pos) - (oldx + oldy) < minimum_pixel) && 
            ((x_pos + y_pos) - (oldx + oldy) > minimum_pixel))) {
                
            return
        }

        oldx = x_pos;
        oldy = y_pos;
        var baseURI = event.target.baseURI;
        var dash_name = baseURI.slice(baseURI.lastIndexOf("/") + 1);
        var d = new Date();

        var args = {};
        args["service"] = "event/fire";
        args["event"] = mouse_event;
        args["x_pos"] = x_pos;
        args["y_pos"] = y_pos;
        args["key_press"] = event.which;
        args["timestamp"] = d.getTime();
        args["widget"] = widget_name;
        args["dashboard"] = dash_name;

        self.call_service(self, args);

    }
```

Hope this clarifies it?

ReneTode:
you can do that already, but not through the callbacks

with a normal jquery selector you can make a callback fro every event you want and do everything there.
i use that for my popup window that closes when clicked anywhere outside the popup as well.

Odianosen25:
Yeah I could have kept it within the widget as you said @ReneTode, but felt its better to allow us open it up within dashboard itself.

This way we could standardise it kind of. Its like every other widgets we have, we could allow them work in silos, but we have them making the callbacks.

I hope I am not missing something?

ReneTode:
then i would suggest changing it to 
domEventData True, instead of event True.

because thats what it makes it possible.
event True is very confusing.

Odianosen25:
Ok @ReneTode,  I do get your point. 

Maybe something like `DOMEvent` instead?

Regards

ReneTode:
its always a DOM event. so i would still use DOMEventData

i would also completely change your doc part (and the description and title from this PR), because what you have given now doesnt describe anything from what it actually does.

what you offer here is:

a choice between
a) the callback recieves no data, or
b) the callback recieves event data.

which probably is obsolete to give that choice, because you can just always provide that info, even when not asked.

and in your docs you say:

> There could be occasions when it is desirable to register for an event other than just "click", and get the whole event data.
This is possible by registering and passing "event" and boolen `true`, so that dashboard is aware of the fact the entire
event data is required. Below is an example

but with your event = true option you can also get the event data from click, it forwards the event data for EVERY action, no matter if its mousemove, click, doubleclick, rightclick, or any other from the dozens of action possibilities.

ReneTode:
thats better odia.

the question still stands if its neccesary to add a choice.
how you did it now it is a choice between 2 functions.

im not familiar enough with jquery to say whats the difference, but why create a difference?
i think its better to change the original function so that it also provides the event data, and then dont provide a choice.

Odianosen25:
I get your point @ReneTode, but that will be @acockburn's call, as I think I am fine with having both.

The original choice allows to just focus on the way AD does it's things, whereby get the entity/event data and do stuffs.

The second will only become useful, if wanting to do more advanced stuffs like your pop-up which I still plan to work towards at some point in the future. 

But as you noticed, we need this to make that to happen and so this is first step.

Regards

ReneTode:
the original choice didnt forward any data at all.
it was just: 
an event happens, then react to it.
opening up all event data for that choice is just a logical choice in my eyes.

but even when a choice is offered, it should be a choice that is between with data, or without. 
and that means it should be
```
if true:
  modified original function, offering data
else:
  original function.
```  
and not:
```
if true:
  a different function, offering data
else:
  original function.
```

Odianosen25:
If I follow what you are saying, wouldn’t that not like literally mean editing all widgets to get the full event data? 😬.

Not something I am really looking forward to doing to be honest. Too much work 😒

ReneTode:
> If I follow what you are saying, wouldn’t that not like literally mean editing all widgets to get the full event data? 

nope.
because in JS you dont need to get the data if offered.

so if the callback is like:

```
function onMouseEvent(self){}
```
it will work the same without any error as
```
function onMouseEvent(self, DOMEventData){}
```
both those functions could be a callback for both options.

in all widgets the first choice is used. so the data is not there, but changing the original function would make the data for all widgets available, when we want to alter it in the future.

for example the switch widget has:
```
    function OnButtonClick(self)
```
for now it can just stay like that. but when someone wants to have eventdata there in the future it could easy be changed to
```
    function OnButtonClick(self, DOMEventData)
```

thats why i suggest not to offer a choice, but just to alter the original function.

Odianosen25:
I just went via the code and I can your your point @ReneTode, will ask @acockburn I will carry out tests locally and if it’s fine then create another PR to make it standard, then see if @acockburn fine with it.

regards

acockburn:
Yes, I agree with Rene’s approach subject to testing. Adding a param doesn’t break anything in JS but makes it available if required, so is a good combination of additional utility and backward compatibility.

PR #1228, opened 2021-03-23
changes in icon widget neccesary for older browsers

ReneTode:


acockburn:
Thanks Rene!

ReneTode:
your welcome

PR #1233, opened 2021-03-26
changes in dashboard from localstorage to cookie, to make old ipads work again

ReneTode:


Odianosen25:
Hello @ReneTode,

Thanks for this. Pls instead of completely removing the localStorage, can we instead test for if the browser supports it?

So we can add a function like below

`
function isLocalStorageAvailable() {

    try {
        var valueToStore = '_storage_test__';
        var mykey = 'appdaemon';
        localStorage.setItem(mykey, valueToStore);
        var recoveredValue = localStorage.getItem(mykey);
        localStorage.removeItem(mykey);

        return recoveredValue === valueToStore;
    } catch(e) {
        return false;
    }
}
`
Gotten from [here](https://stackoverflow.com/a/46581054)

So isLocalStorageAvailable will essentially let us know if you can use it. If we can, we make use of localStorage instead. If not, then use cookie.

So we will create a separate function that gets and sets the deviceid, within which we will carry out the checks before setting it retrieving the deviceid.

Also I think we should stick to the precious one given as `hadashboard-deviceid `, whether using localStorage or not to allow for uniformity.

Hope it makes sense?

Kind regards

Odianosen25:
But looking at the original code again, there is already a way to cater for if there is no local storage, so not sure how this resolves it. Or is it possible the error is not being captured [here](https://github.com/AppDaemon/appdaemon/blob/dev/appdaemon/assets/templates/dashinit.jinja2#L63)?

ReneTode:
localstorage just gives trouble, even when its between try/catch.
and i decided to change the cookie name, because of the name Andrew already used and to keep it short.

PR #1236, opened 2021-03-28
Updated Change Log

Odianosen25:


ReneTode:
description is wrong ;)
because it fixed the issue for all devices (devicid was not properly working at all)
and it fixed the issue that the device code blocked old IOS devices.

PR #1241, opened 2021-04-06
Allow to disable Widget assume state for Lights and Switches

Odianosen25:
This PR allows for an option `simulate_feedback`. Whereby the feedback of the widget, is solely based on the feedback of the entity.

When activated by having `simulate_feedback: 0` in the widget config, this removes the possibility of the widget having the wrong feedback in the event the service call didn't actually control the device properly. And it leads to dashboard being out of sync and therefore kind of unusable until the device is manually set to the correct state as the widget.


ReneTode:
sorry odia, i really dont understand this.

the dashboard always gives the right state from the entity, but you want the light and switch widget to able to change without changing the state??

the title from this PR is saying nothing (please add dashboard and which widgets)

and i think you are correcting something on the wrong place.
dashboard always gives the identical state as AD (if that is sometimes wrong there is a bug somewhere)
and it should always give that state.

Odianosen25:
@ReneTode,

I have updated the title as requested.

On dashboard, I noticed that when its clicked, dashboard auto updates the state of the entity even before the service call is completed as [here](https://github.com/AppDaemon/appdaemon/blob/dev/appdaemon/widgets/baselight/baselight.js#L125). What happens as I had tried to explain, is that sometimes if the service call for some reason is broken, even if the entity's state is not the required state, dashboard will assume its the state that was requested. And therefore dashboard simulates the feedback of the entity.

As I had a custom system using this (not via HA), I saw myself in a situation whereby the state on dashboard was out of sync with the physical device I was controlling over mqtt. 

Hope it makes sense now?

ReneTode:
> As I had a custom system using this (not via HA), I saw myself in a situation whereby the state on dashboard was out of sync with the physical device I was controlling over mqtt.

if that happens the state from the AD entity is also out of sync with the physical device.

with this you disable the change from the view, so the icon isnt updated at all.
it just reacts to state changes.

i dont understand you argname chose from that point of view.
it has nothing to do with simulating or with feedback.

the reason why andrew did set it up like this is speed.

what it does:
click button > send service > assume it goes right > state changes > dashboard is already correct.

if you want to add this feature then i would call it something like: 
assume_state: 1
where assume state is default.

and in that case add it to all widgets that change state from entities, not just these 2.

Odianosen25:
> with this you disable the change from the view, so the icon isnt updated at all.
it just reacts to state changes.

To an extent yes. It only changes on state change only

> the reason why andrew did set it up like this is speed.

I understand that. But even by relying on the state alone, its pretty quick from my tests and I didn't really notice any significant change in speed. Maybe before it mattered, but with the new updates, its quite fast.

> assume_state: 1

I don't mind

> and in that case add it to all widgets that change state from entities, not just these 2.

I was afraid you would say that. I had done it for only light (which was my target), then added switch as I felt you will pick up on it. Hoping you will just allow me pass with that alone, but I guess my trick didn't work lols

> and add to the docs that this will cause delayed updating from the state in dashboard

its not really slow trust me. But yeah I can add that too.

ReneTode:
> its not really slow trust me. But yeah I can add that too.

that really depends on the environment you are in. for some of my dashboards it would be slow.
(dont forget to update the title here, if you change it)

but all that said, i still think you correct it on the wrong place.
your app should take care of services that go wrong.

Odianosen25:
> i asked for it because i also had switches that didnt always go right. i corrected that in the apps, because you want your light to go on always when you press the button. not just sometimes.

Yeah I get it. But I use zigbee, and sometimes it beyond my control that it doesn't follow suit as I had hoped.

I will try and work on the changes and get back. But really can't we just allow for the light and switches to go in for now? Kind are lots of work to update all. Then those that need it for others, can be added over time

ReneTode:
> But really can't we just allow for the light and switches to go in for now? Kind are lots of work to update all. Then those that need it for others, can be added over time

i would say no.
you can use a change like this for yourself in custom widgets, but i dont think its smart to add such a function incomplete. and it is an incomplete function when its only available for 2 widgets.

my personal opinion is: add all or add none if it is a general function (which this actually is)

if you add it only for 2 widgets, the others will get forgotten, and widgets will become more and more different from each other, even in functions that are relativ general.

maybe its even better to think about a general way to implement it (and the set_view as well)

acockburn:
closing

PR #1257, opened 2021-05-01
Fix issue with state

Odianosen25:


ReneTode:
as i see it an entity should always have a state.
as soon as it is created it should get a state, that state could be None, "unset", "unavailable" or anything you like, but a state should exist.

if there is a way that an entity is created without a state, then that should be corrected, not the listen_state.

Odianosen25:
Hello @ReneTode,

I do agree, but when fixing an issue with the `attribute`, so its doesn't break, this came up
"None/""" or anything can kind are be same. I am not even sure what leads to this, but just put in a fix for it for now, as its better its ignored than giving a user an error in the logs.

It doesn't affect what AD does underneath really, and so better no error. In the Admin it comes as `undefined`.

ReneTode:
> as its better its ignored than giving a user an error in the logs.

oh no. if there is a problem, a bug, or whatever it isnt good to ignore it at all.
it should give an error for sure if the entity doesnt have a state. because an entity should always have a state.
so thee is a bug somewhere in the place where the entity is created (create_entity or plugin)

ignoring that could lead to other problems.

Odianosen25:
Also I forgot to add, this only affects when listen_state is used with the “immediate” flag. So it really doesn’t affect anything. 

Other aspects works as fine, as there are checks for if the entity exists (which I had fixed some while back), as this fixes the state.

ReneTode:
its an obsolete check. code that is unneeded.
because when the entity exists (which should be checked) then you need no check if the state exists, because it should always exist.

Odianosen25:
> its an obsolete check. code that is unneeded.
because when the entity exists (which should be checked) then you need no check if the state exists, because it should always exist.

Ok so should I give some form of warning or something? So the user knows there is something wrong?

It is possible unfortunately in AD to create an entity without a state, using the `set_state` api. So if the user encounters this, I guess giving a warning should suffice?

Regards

ReneTode:
> It is possible unfortunately in AD to create an entity without a state, using the set_state api.

so thats where the error is. state in set_state shouldnt be optional if the entity doesnt exist.
that should be corrected, and at that place should be a warning like:
"you are trying to create an entity without a state, AD did not make the entity."

if that is corrected, then the warning you want to place here would never appear.

PR #1262, opened 2021-05-04
Fix typo in `app/disable` service

engrbm87:
Fix typo in `app/disable` service documentation.


Odianosen25:
@engrbm87,

Thanks for this, but any reasons why this fails? Did you carry out a pre-commit check?

Kind regards

engrbm87:
Not sure what is the cause of the error. I made the edit on github webpage so I don't think I can run pre-commit check..

PR #1266, opened 2021-05-10
Added !include

Odianosen25:
This PR allows to make use of the `!include` tag in the configuration files

it also contains a fix for the pre-commit. As the dashboard setup was causing the checks to fail due to the recent addition of the `aui`. So to fix this, had to carryout a little refactor and apologies I added it to the same PR.

### DIFF
--- a/appdaemon/__main__.py
+++ b/appdaemon/__main__.py
@@ -239,6 +239,12 @@ class ADMain:
         try:
 
             #
+            # Read config file using include directory
+            #
+
+            yaml.add_constructor("!include", utils._include_yaml, Loader=yaml.SafeLoader)
+
+            #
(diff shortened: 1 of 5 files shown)


ReneTode:
and you did a background refactor from dashboard.

i dont think those 2 things belong in the same PR, or you should at least make it clear in title and description ;)

Odianosen25:
> and you did a background refactor from dashboard.

Yeah I could not pass checks, as it was @acockburn that caused it lols

>  dont think those 2 things belong in the same PR, or you should at least make it clear in title and description ;)

I agree, but just felt well it was extra work and I worked off my main dev branch locally unfortunately. I will give more description as said.

PR #1282, opened 2021-06-02
Plugin restart fix

Odianosen25:
This fixes an issue where AD crashes and needs a manual restart when the following happens:

- Multiple plugins (>=3) are being used
- Multiple apps have no `plugin` directive, so they will restart when any plugin re-initialises
- All the plugins restarts at the same time; this can be due to network going down and then up


ReneTode:
2x the same PR?

Odianosen25:
@ReneTode wanted to make a change to the PR, but my Git server was broken, so for some reasons things got messed up.

So closed the previous one and re-submitted

PR #1311, opened 2021-07-24
widget media_player: corrected state passing

Tigger2014:


ReneTode:
@Odianosen25 this corrects a problem that somehow got into the widget.

PR #1314, opened 2021-07-29
Addition: fan widget

Tigger2014:


ReneTode:
@Odianosen25 please accept this new widget.
its developed together with me and i think its a good addition to the widgets.

PR #1364, opened 2021-09-28
Fixing hadashboard on old IOS devices

clyra:
It seems that the old IOS devices doesnt support try/catch statements and there's one on basejavascript. This PR removes that try/catch and now the dev branch is working on my ipad2/ios 9.3.5.


acockburn:
That is the most helpful thing that has happened all year - thanks!!

Issue #1366, opened 2021-09-30
listen_state callback is not firing with new=0


deviantintegral:
I was tracing why an app wasn't working, and I got down to this fragment:

```python
class WashingMachine(hass.Hass):
    def initialize(self):
        self.log("Starting WashingMachine...")
        self.listen_state(self.notify_done, entity="sensor.washing_machine_smartenergy_metering", new=0.0,
                          immediate=True)

    def notify_done(self, entity, attribute, old, new, kwargs):
        self.log(entity + " changed to " + str(new) + "W")
```

I would expect that this would immediately log if the state is `0` or `0.0`. However, the callback is never called.

If I remove the `new` parameter, it is... but `new` in the callback shows as 0.0:

```
2021-09-29 21:48:27.418317 INFO AppDaemon: Initializing app laundry_room_washer using class WashingMachine from module laundry_room
2021-09-29 21:48:27.434270 INFO laundry_room_washer: Starting WashingMachine...
2021-09-29 21:48:27.440444 INFO laundry_room_washer: sensor.washing_machine_smartenergy_metering changed to 0.0W
```

It doesn't appear that there's any debug logging in `appdaemon.state.State.add_state_callback()` that would help track this down, but a first read doesn't show any obvious bugs. My next step would be to hack in some additional logs, but if there's something I've missed I'd appreciate a pointer.

For now, I can work around this by returning early in the callback.

deviantintegral:
I went to do this, but then got the following hint in the logs:

> TypeError: '>' not supported between instances of 'str' and 'int'

So, it looks like the value is coming in as a string, and not a numeric type. Changing to `new="0.0"` causes the callback to be executed.

This was the first automation where I was attempting to interact with a number. I suppose all states in HA are treated as strings? If so, I wonder if the docs at https://appdaemon.readthedocs.io/en/latest/AD_API_REFERENCE.html#appdaemon.adapi.ADAPI.listen_state could be updated to call that out.

Justihar:
Correct, the value for "new" from an HA entity is always going to be a string but be careful because when using immediate=True, the old value is going to be a None object on the first callback run. Also, when you first start the app, if the washing machine value is 0, your application will think it has stopped even if it wasn't previously running. Another option would be to omit the new parameter from the callback and use something like this in the callback method.

```python
try:
    new = float(new)
except (TypeError, ValueError):
    new = 0.0

if old is None:     # immediate=True
    if new > 0:
        # app just started and machine is already running
[...]

PR #1370, opened 2021-10-10
Document old and new callback parameters are strings

deviantintegral:
Following up from https://github.com/AppDaemon/appdaemon/issues/1366, this makes it explicit that the old and new parameters are always strings, as that's how HomeAssistant stores them.

### DIFF
--- a/appdaemon/adapi.py
+++ b/appdaemon/adapi.py
@@ -1272,10 +1272,10 @@ class ADAPI:
                 state change within the specified entity, and supply the callback functions with
                 the entire state dictionary for the specified entity rather than an individual
                 attribute value.
-            new (str, optional): If ``new`` is supplied as a parameter, callbacks will only be made if the
+            new (optional): If ``new`` is supplied as a parameter, callbacks will only be made if the
                 state of the selected attribute (usually state) in the new state match the value
                 of ``new``.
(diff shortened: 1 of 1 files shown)


deviantintegral:
The failures look to be in an unrelated JS file: https://github.com/AppDaemon/appdaemon/pull/1370

ReneTode:
this shouldnt be changed in the docs, because old and new can be integers as well when the entity is in a different namespace then HA.

deviantintegral:
Ah, I didn't realize this method was used outside of Home Assistant. Would it be OK to note that the values are strings for Home Assistant specifically in the parameter description?

deviantintegral:
It figures that attributes allow more types - it would explain why I was surprised to see a float as a string, when I was accessing attributes before that were "properly" typed.

Where this leads me is to suggest a _warning_ to be careful on types. After all, an addon could create an entity that stores an int as a string even when it's not required, and the design of HA means that we can't be stricter.

One minute and I will propose a suggestion.

Odianosen25:
So @ReneTode are we good with this change? So I can merge it

ReneTode:
i said so didnt i? ;)

Odianosen25:
Ah ok then thanks

PR #1477, opened 2022-02-10
Allow for reloading of app's module when a global module dependency reloads

Odianosen25:
In AD, when an app's `global_dependency` module reloads, the app's module doesn't automatically reload
This many a times is not an issue, if the module is used or instantiated within the `initialize` function.
But if the module's function is used outside the app's class like as a decorator or standalone function to be passed to another processor, the app never gets access to the latest update since the app's module never reloaded.

So this PR will ensure that when the `global_module` reloads, so also does the app's module reload.
This will also ensure that other apps, though using the same module that got reloaded will not restart and will continue to function. But they will only get the latest update, when the apps reloads themselves

### DIFF
--- a/appdaemon/app_management.py
+++ b/appdaemon/app_management.py
@@ -755,9 +755,10 @@ class AppManagement:
 
     # noinspection PyBroadException
     # Run in executor
-    def read_app(self, file, reload=False):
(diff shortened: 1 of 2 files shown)


acockburn (changes requested):
See my comments in the main thread

acockburn:
closing

Issue #1492, opened 2022-03-23
Need help with register_route


foxy82:
Hi,

I'm trying to use register route to serve a HTML page.

I have the following:

```
class my_app(hass.Hass):
    def initialize(self):
        self.set_log_level("DEBUG")
        self.register_route(self.auth_call, "foobar")

    async def auth_call(self, request, kwargs):
      self.log(request)
      return "Hello World", 200
```

 I request: http://<IP:PORT>/app/foobar

In the logs I see:
```
2022-03-23 10:21:10.043486 INFO my_app: <Request GET /app/foobar >
```

In Chrome I get back:

This page isn’t working
<IP> didn’t send any data.
ERR_EMPTY_RESPONSE

Not sure if I'm returning the wrong thing. The documentation for `register_route` reads:

> It should be noted that the register function, should return a string (can be empty), and an HTTP OK status response (e.g., 200. If this is not added as a returned response, the function will generate an error each time it is processed.

Which I appear to be doing.

AppDaemon version 4.2.0.

In case it matters I'm actually trying to implement an Alexa Smart Home skill but to do that I need to do OAuth. I do that using Login With Amazon (LWA) but I'm not receiving the AuthorizeGrant message but I can see some other http requests coming in hence trying to get a page up.

Justihar:
Try using a Response object from aiohttp.web.

```
from aiohttp import web

class my_app(hass.Hass):
    def initialize(self):
        self.set_log_level("DEBUG")
        self.register_route(self.auth_call, "foobar")

    async def auth_call(self, request, kwargs):
      self.log(request)
      return web.Response(text="test", content_type="text/html")
```

foxy82:
Thanks that did it.

foxy82:
Create PR to change the docs.

PR #1493, opened 2022-03-23
Fix register_route documentation

foxy82:
Fix `register_route` document which needs to return an aiohttp Response object.

As per the response in this issue: https://github.com/AppDaemon/appdaemon/issues/1492


Odianosen25:
Hello @foxy82,

Please can you run `pre-commit run --all-files` on your PR so it passes its checks?

PR #1496, opened 2022-03-30
Container improvements

NigelRook:
Provides two major benefits:
- Reduced image size to ~285Mb
- Now compatible with alpine's py3-* packages

Has the downside that this is potentially a breaking change for some people - the final image no longer contains rust/gcc, so if anybody is running any additional requirements that need those, they'd need to add them back through system_requirements.txt. This is mitigated somewhat by the fact that they might just be able to add an alpine package for it (I did this work for scikit-learn, which takes ~20 minutes to compile on my low-power i3 server, vs 20 seconds to install the alpine package + deps). That still requires a config change for people though, so I understand if you won't want to merge this. I'd be willing to add those packages back in to this PR if you'd prefer.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,7 +4,15 @@
 Dockerfile
 .dockerignore
 
+azure-pipelines.yml
+.pre-commit-config.yaml
+.semaphore
+
 *.md
 !README*.md
+MANIFEST.in
+setup.py
 *.rst
+
 docs
+conf
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,6 @@
 .git
 .gitignore
+.github
 
 Dockerfile
 .dockerignore
@@ -16,3 +17,4 @@ setup.py
 
 docs
 conf
+.readthedocs.yml
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,7 +29,9 @@ VOLUME /certs
 WORKDIR /usr/src/app
 
 # Install runtime required packages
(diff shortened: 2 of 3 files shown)


acockburn:
I am worried about introducing breaking changes, especially since I don't use docker myself.

acockburn:
If you can make the change to include alpine vs compile it I think that is a good compromise

NigelRook:
OK updated

For clarity, I've:

- rebased onto latest dev
- added all alpine packages which were present in the current dev image
- bumped alpine to the latest stable release

Image isn't much smaller anymore, but alpine py3-* packages are still working.

acockburn:
Thanks!

acockburn:
I had to revert these commits because my docker pipeline started failing - not sure if this is the cause or not

acockburn:
If/when I fix the build, I'd like to try again with this PR but it's merged and reverted and I am not sure there is a way I can re-merge it - could I ask you to create a fresh PR for this please?

Also, it seems your theory about memory and the size of the image was correct - I managed to get a build with one image, so now I am trying for 3. If that works I will call it good and drop support for linux/arm/v6. I have no idea if that is the right one to drop as I don't know who is even using AD ;) But I guess that the oldest non-intel image is probably the least used.

NigelRook:
Here you go: https://github.com/AppDaemon/appdaemon/pull/1655

Rebased and tested on my end

acockburn:
Thanks!

PR #1515, opened 2022-06-02
Correct docs

illuzn:
The references to 1 should be 0 here to supress the warnings


Odianosen25:
Hello @illuzn,

Thanks for the PR but can you redo it against the dev branch instead?

PR #1522, opened 2022-06-25
Update basetemperature.js

DaveDixon:
When the temperature widget is first displayed, the red bar is always at the minimum temperature, not the current temperature. If the page is left alone long enough, it updates to the current temperature. But navigation to another page and returning causes it to reset to the minimum temperature. This change causes it to display the red bar at the current temperature even the first time. This is all assuming that the call to self.gauge.update(self.parameters.settings) is correct here.


Odianosen25:
Will be deleting this @DaveDixon, since new one summitted.

Regards

PR #1540, opened 2022-06-29
Added decorator constraint

Odianosen25:
This adds a feature of enabling the use of decorators for constraints


acockburn:
As discussed - we think this is a good idea, but need to figure out a way to prevent the timers from being inaccurate. Maybe detect if the decorator is being called vs an actual callback, and if it is the decorator, don't increment the counters in the core, then increment them in the decorator only if the callback is actually called.

PR #1556, opened 2022-08-08
Reduce installation requirements

gvalkov:
The azure dependencies don't seem to be used anywhere or at least I can't find any references. Pygmentize and wheel are also only needed during development. Pygments can probably be just pulled through Sphinx ... don't know what is gained by pinning it here. 

This removes `~175MB` from the installation size (not accounting for any overlap with the other installation dependencies). 

```
$ python3 -m pip install -t . azure-mgmt-compute==27.2.0 azure-mgmt-storage==20.0.0 azure-mgmt-resource==21.1.0 azure-storage-blob==12.12.0 azure-keyvault-secrets==4.4.0 pygments==2.12.0 wheel==0.37.1
$ du -hs
175MB
```


acockburn:
These libs are used for the deployment pipeline so I don't think we can remove them.

mion00:
Thanks for your suggestion!
On the `dev` branch we have drastically improved how we package the application and handle its dependencies, separating between the ones needed for development, building docs and runtime.
See #1673 for reference.

PR #1583, opened 2022-10-12
refactor repeated SSL/auth cruft for HASS connection

zanfur:
This pulls all the copypasta auth and ssl handling into a single place for the aiohttp session, and as a bonus implements the todo around using the cert_path argument.  Additionally, put the ha_url into the session's base_url rather than prepending it everywhere.

With this change, now any appdaemon app can use `plugin.session.request("get", "/api/whatever")` to make arbitrary HA rest calls through the AD session, allowing full access to HA data even if there's no integrated AD support for it (like Devices or Areas or search).

### DIFF
--- a/appdaemon/plugins/hass/hassplugin.py
+++ b/appdaemon/plugins/hass/hassplugin.py
@@ -25,9 +25,7 @@ def hass_check(func):
     def func_wrapper(*args, **kwargs):
         self = args[0]
         if not self.reading_messages:
-            self.logger.warning(
-                "Attempt to call Home Assistant while disconnected: %s", func.__name__
-            )
+            self.logger.warning("Attempt to call Home Assistant while disconnected: %s", func.__name__)
             return no_func()
         else:
             return func(*args, **kwargs)
@@ -48,13 +46,9 @@ class HassPlugin(PluginBase):
 
         # validate basic config
         if "ha_key" in args:
-            self.logger.warning(
-                "ha_key is deprecated please use HASS Long Lived Tokens instead"
-            )
+            self.logger.warning("ha_key is deprecated please use HASS Long Lived Tokens instead")
         if "ha_url" not in args:
-            self.logger.warning(
-                "ha_url not found in HASS configuration - module not initialized"
-            )
+            self.logger.warning("ha_url not found in HASS configuration - module not initialized")
 
         # Locally store common args and their defaults
         self.appdaemon_startup_conditions = args.get("appdaemon_startup_conditions", {})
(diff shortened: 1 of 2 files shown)


zanfur:
Looks like I need to black format, will do that and update in a couple hours.

appdaemon/plugins/hass/hassplugin.py
```diff
@@ -446,24 +442,14 @@ def utility(self):
...
         self.logger.debug("set_plugin_state() %s %s %s", namespace, entity_id, kwargs)
-        config = (await self.AD.plugins.get_plugin_object(namespace)).config
 
-        # TODO cert_path is not used
-        if "cert_path" in config:
-            cert_path = config["cert_path"]
-        else:
-            cert_path = False  # noqa: F841
+        # if we get a request for not our namespace something has gone very wrong
+        assert namespace == self.namespace
```
zanfur, line 447: I left these as an assert because iiuc we should really never get a mismatch or that indicates a serious logic error.  If there's a more appropriate exception to throw I'm happy to use another.

appdaemon/plugins/hass/hassplugin.py
```diff
@@ -42,76 +42,41 @@ def __init__(self, ad: AppDaemon, name, args):
...
-            self.plugin_startup_conditions = args["plugin_startup_conditions"]
-        else:
-            self.plugin_startup_conditions = {}
+        # Locally store common args and their defaults
+        self.appdaemon_startup_conditions = args.get("appdaemon_startup_conditions", {})
+        self.cert_path = args.get("cert_path")
+        self.cert_verify = args.get("cert_verify")
+        self.commtype = args.get("commtype", "WS")
+        self.ha_key = args.get("ha_key")
+        self.ha_url = args.get("ha_url", "")
```
zanfur, line 59: This is a behavior change.  Previously, if you didn't specify the ha_url, it would crash on startup with "no attribute 'ha_url'".  Now it will fail to connect instead.

Odianosen25:
Ok tested the PR @zanfur and from initial tests looks ok. But I and @acockburn need to be sure of some stuffs before it can be merged as @acockburn need to give the final sign-off anyway. The first thing will be to avoid the exposure of the ws object as its not a good idea to allow for users to be able to create new ones at will

zanfur:
This PR doesn't change anything about how exposed the ws object is, just refactors how it's created. It's still just a member variable on the plugin object. I considered not even saving it there, because it isn't actually necessary, but that's how it was previously so I left it.

acockburn:
I don't see any immediate problems here, if @Odianosen25 says it tests out OK I am happy to merge it into dev for further testing

Issue #505, opened 2019-01-31
Add support for Area and Device registries


Kane610:
To continue with being on par with Home assistants API there should be support in Appdaemon for the Area and Device registries.

Odianosen25:
@Kane610,

Thanks for this. Can you kindly explain more on what you mean?

Regards

Kane610:
Devices is the physical representation of one or more entities in home assistant.
Areas are physical representation of placement of a device.

Would be nice to be able to have helper functions that can be used with areas and devices

ReneTode:
areas has been merged 3 days ago.
it isnt even in the latest version. and i cant even find a trace from it in the docs.
i am sure that at some point there will come helper functions for it, but its very hard to implement something that we dont even work with.

i think that when people are actually working with them, there will be suggestions on how to use and what kind of helper functions could be usefull. thats about the time to work on implementing it.

acockburn:
I agree with Rene - when it's clear what we can add that will help in this area we can look into it.

ReneTode:
i dont know what MVP support means, but i am sure Andrew is willing to look at a PR from you if you think it is that important.

Kane610:
Minimum viable product, something to start with to evaluate further use cases

ReneTode:
ah, ok. thanks for explaining.

I think that in case from something as new as this in HA, its always wise to let it go on HA for a while.
let people see what to do with it, if it is buggy and if HA will make breaking changes (as they almost always do)

at this moment we are preparing AD for beta release for AD 4.0 and then its not wise to add something new from HA, that they didnt even release (and from experience its in general not wise to support something new from HA when its not at least 2 releases old) ;) 

but like Andrew said, he will add it at some point and he marked it as enhancement.

unconfigured:
Nice to see there is already a request for it. I would like to use areas to represent rooms and I have just searched for a way to lookup to which area/room a motion sensor belongs to than query other devices in that area/room (lights, media_player, ...) and react accordingly. This would be much nicer than maintaining groups with all devices for all rooms. Cycling through entities and do some magic is very powerfull and convenient in appdaemon - thanks for this! I would love to see area support and just wanted to signal, that there is a demand an use-cases for that :)

ReneTode:
whats the advantage from an area called livingroom over a group called livingroom?
in both cases you need to ad the devices that belong to livingroom.

i have groups like that since 2 years, i like to know what advantage it would give to change to areas?

[...]

PR #1584, opened 2022-10-15
Plugin registry support

zanfur:
This currently only implements in the plugin, still need to do the API plumbing for it, but wanted to put it up for early feedback in case I'm doing something stupid already.

### DIFF
--- a/appdaemon/plugins/hass/hassapi.py
+++ b/appdaemon/plugins/hass/hassapi.py
@@ -2,6 +2,7 @@ from typing import Any, Optional
 import requests
 from ast import literal_eval
 from functools import wraps
+from copy import deepcopy
 
 import appdaemon.adbase as adbase
 import appdaemon.adapi as adapi
@@ -738,3 +739,30 @@ class Hass(adbase.ADBase, adapi.ADAPI):
             return literal_eval(result)
         except (SyntaxError, ValueError):
             return result
+
+    #
+    # Helper functions for HASS-specific functionality
+    #
+
+    @utils.sync_wrapper
+    async def hass_get_registry(self, registry, id=None, copy=True, **kwargs):
+        ns = self._get_namespace(**kwargs)
+        plugin = await self.AD.plugins.get_plugin_object(ns)
+        entries = plugin.registry[registry]
+        result = entries[id] if id else entries
+        return deepcopy(result) if copy else result
+
+    def hass_get_area(self, namespace, area_id=None, **kwargs):
+        return self.hass_get_registry("area", id=area_id, **kwargs)
+
+    def hass_get_device(self, device_id=None, **kwargs):
+        return self.hass_get_registry("device", id=device_id, **kwargs)
+
+    def hass_get_entity(self, entity_id=None, **kwargs):
+        return self.hass_get_registry("entity", id=entity_id, **kwargs)
+
+    @utils.sync_wrapper
+    async def hass_search_related(self, registry, id, **kwargs):
+        ns = self._get_namespace(**kwargs)
+        plugin = await self.AD.plugins.get_plugin_object(ns)
+        return await plugin.search_hass_registry(registry, id)    
--- a/appdaemon/plugins/hass/hassapi.py
+++ b/appdaemon/plugins/hass/hassapi.py
@@ -767,4 +767,4 @@ class Hass(adbase.ADBase, adapi.ADAPI):
     async def hass_search_related(self, registry, id, **kwargs):
         ns = self._get_namespace(**kwargs)
         plugin = await self.AD.plugins.get_plugin_object(ns)
-        return await plugin.search_hass_registry(registry, id)    
+        return await plugin.search_hass_registry(registry, id)
--- a/appdaemon/plugins/hass/hassapi.py
+++ b/appdaemon/plugins/hass/hassapi.py
@@ -745,6 +745,7 @@ class Hass(adbase.ADBase, adapi.ADAPI):
     #
 
     @utils.sync_wrapper
+    @hass_check
     async def hass_get_registry(self, registry, id=None, copy=True, **kwargs):
         ns = self._get_namespace(**kwargs)
         plugin = await self.AD.plugins.get_plugin_object(ns)
@@ -762,6 +763,7 @@ class Hass(adbase.ADBase, adapi.ADAPI):
         return self.hass_get_registry("entity", id=entity_id, **kwargs)
 
     @utils.sync_wrapper
+    @hass_check
     async def hass_search_related(self, registry, id, **kwargs):
         ns = self._get_namespace(**kwargs)
         plugin = await self.AD.plugins.get_plugin_object(ns)
--- a/appdaemon/plugins/hass/hassapi.py
+++ b/appdaemon/plugins/hass/hassapi.py
@@ -753,7 +753,7 @@ class Hass(adbase.ADBase, adapi.ADAPI):
         result = entries[id] if id else entries
         return deepcopy(result) if copy else result
 
-    def hass_get_area(self, namespace, area_id=None, **kwargs):
+    def hass_get_area(self, area_id=None, **kwargs):
         return self.hass_get_registry("area", id=area_id, **kwargs)
 
     def hass_get_device(self, device_id=None, **kwargs):
--- a/appdaemon/plugins/hass/hassplugin.py
+++ b/appdaemon/plugins/hass/hassplugin.py
@@ -96,7 +96,7 @@ class HassPlugin(PluginBase):
         return []
 
     #
-    # Persistent Session to HASS instance
+    # Persistent HTTP Session to HASS instance
     #
     @property
     def session(self):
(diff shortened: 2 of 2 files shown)


acockburn:
Took a look and I am mostly happy with it - pretty much what I was hoping for, code is nice and concise. The only comment I have is that in other parts of AD we use UUIDs rather than random numbers for unique identifiers. Could you change this to a use a UUID for consistency please? Then we should be able to merge and test further. You can find examples of UUID use in for instance events.py, it looks like this:

```
id = uuid.uuid4().hex
```

zanfur:
This is a websocket id, and must be a number.  Trying to use a string fails:

```
2022-10-17 13:12:15.957325 WARNING HASS: response: `{'id': '97e896950b054e378122c15e1b32b45c', 'type': 'result', 'success': False, 'error': {'code': 'invalid_format', 'message': 'Message incorrectly formatted.'}}`
```

zanfur:
From RFC6455 (WebSocket):

   The masking key is a 32-bit value chosen at random by the client.
   When preparing a masked frame, the client MUST pick a fresh masking
   key from the set of allowed 32-bit values.  The masking key needs to
   be unpredictable; thus, the masking key MUST be derived from a strong
   source of entropy, and the masking key for a given frame MUST NOT
   make it simple for a server/proxy to predict the masking key for a
   subsequent frame.  The unpredictability of the masking key is
   essential to prevent authors of malicious applications from selecting
   the bytes that appear on the wire.  [RFC 4086](https://datatracker.ietf.org/doc/html/rfc4086) [[RFC4086](https://datatracker.ietf.org/doc/html/rfc4086)] discusses what
   entails a suitable source of entropy for security-sensitive
   applications.

zanfur:
This should completely close #505 when merged.

Odianosen25:
Hello @zanfur,

Thanks for the PR. From my end, can you kindly remove the word “hass” in the function names? There is no point in it being there. So instead of “get_hass_registery”, should just say “get_registry. As since it’s just in hass object, no point on it being there; I made this mistake when developing the mqtt api 😉.

Also can these api calls be made into service calls? So the plugin handles it as a service, and in hassapi, it’s simply making a call to the service under the hood. This makes it easily accessible by other apps running in another namespace that needs the data without need or calling another app. Just ensure in hassapi, it has “return_result” set to true.

Lastly I am still not a fan of creating and closing a ws connection at will when needed. It’s like using ws, as a rest protocol and defeats it’s purpose. I just feel we might be getting ourselves into an issue we not seeing yet on the side of HA. Can this be modified pls?

kind regards

zanfur:
No problem on changing the names, I didn't know which way you guys wanted.
Though if it's a service instead anyway, what would the service name be?
The issue with making it a service is that all the other services in this
namespace are actual hass services; this would be the odd one out. Should
it have its own namespace?

To re-use the ws would be a much more invasive change, involving
substantially more complicated code than exists in the entire plugin
currently, because of the possibility of running multiple callbacks in
parallel and having them collide.  It basically requires writing an async
framework.  Also, the connection setup is nearly instant (it's just http
after all) and the results are the entire registry, which make the
connection setup a drop in the bucket anyway. I don't know that it's worth
it for this. If there *was* a REST api it would make sense to use that
instead, there just isn't one for this. The registries just aren't updated
very frequently.

Also this PR in particular is agnostic to whether the ws is used or not;
that can be implemented or not and ws_request would still have the same
api. The previous (already merged) PR is what made the basic ws_request
functionality. I can iterate on that to make the ws be reused, but I think
that would be in a separate PR unrelated to this one that just exposes the
existing functionality.

Odianosen25:
@zanfur 
> I didn't know which way you guys wanted.

Should be mostly on the `hassapi` module file. The plugin one could be ignored

> Though if it's a service instead anyway, what would the service name be?

It will be something in the likes `registry/get`, `registry/search` and `areas/get`

> The issue with making it a service is that all the other services in this
namespace are actual hass services; this would be the odd one out. Should
it have its own namespace?

No that will not be the case as we do have like 2 or so services that are not available in HA and we manually added them [here](https://github.com/zanfur/appdaemon/blob/dev/appdaemon/plugins/hass/hassplugin.py#L779)

> To re-use the ws would be a much more invasive change, involving
substantially more complicated code than exists in the entire plugin
currently, because of the possibility of running multiple callbacks in
parallel and having them collide.

Agree its a bit more complicated, but its the right thing to do as it will also make the call faster compared to setting up and closing a new connection each time. There is no fear of multiple callbacks, as if its made a service call, AD internally guarantees a single service at every point in time, added to the ws client is single threaded. 

> I can iterate on that to make the ws be reused, but I think
that would be in a separate PR unrelated to this one that just exposes the
existing functionality.

I don't think we should separate it, or if at all we should, the re-use should come in first making it agnostic then we build on that to implement this or any other functionality on getting extra data from HA over if that makes sense.

Odianosen25:
Following up from the above @zanfur, the following code could replace `ws_request`

```python

def init(self):

    self._ws_results = {}

async def ws_request(self, api, **kwargs):
        id = random.getrandbits(64)
        args = json.dumps(
            {
                "id": id,
                "type": api,
                **kwargs,
            }
        )
		
		_event = asyncio.Event()
        self._ws_results[id] = {"event": _event} 
        await utils.run_in_executor(self, ws.send, args)
        await _event.wait()		
        response = self._ws_results.pop(id)
        return response["response"]

async def get_updates(self):  # noqa: C901
		# stuffs
        
		if result["id"] in self._ws_results:
    		# results to be picked up
			self._ws_results[result["id"]] = {"response": result["result"]}
			self._ws_results["event"].set()
			continue

        elif not (result["id"] == _id and result["type"] == "event"):
            self.logger.warning("Unexpected result from Home Assistant, id = %s", _id)
            self.logger.warning(result)

```
Something like that

zanfur:
I'll play with that. I was going to use a promise/future but 5 yes is the
same. If there's no chance of multiple threads calling ws_request in
parallel then agree it's much simpler.

Odianosen25:
@zanfur 

> Event is not thread safe, you can't safely use it how your example code does.

Can you plain a bit more pls. As once you create it, its only `set` that needs being releasing from another thread which is fine. I use it quite heavily in a few apps myself

> We can't add new subscriptions to the same id, if we ever want another.

If that is the case, then we should add the setting to the HA config, so the user should set it up at start up. If its absolutely we cannot, then we can also manage it by only setting up the subscriptions at start-up, then we have an absolute number of ids we using for it and place the logic based on that.

Odianosen25:
The event itself is not be to used across threads, you still waiting in the same function.

zanfur:
From the asyncio.Event documentation (https://docs.python.org/3/library/asyncio-sync.html) at the top of the page:

> asyncio primitives are not thread-safe, therefore they should not be used for OS thread synchronization (use threading for that);

It's one of those things that will work, usually, until it explodes. You need to use concurrency.Event for cross-thread synchronization. If the request is being run from an executor somewhere, it will be notified across threads (from the loop doing the receive to the thread waiting for the result).

Odianosen25:
Hello @zanfur,

Yeah I get you point. But we not (at least as I understand it), working across threads on this. The the event object is not expected to be accessed within an executor, but pls let me know if I am getting anything mixed up in my head. 

The two functions being referred to here `get_updates` and `ws_requests` which are both async will run within the same loop. So not sure how we get into a different thread at this point really.

zanfur:
From my analysis of the code (which hey could easily be wrong), the app gets its own thread and all its calls to the plugin (via the api) are run in that thread's loop (or the executor if it's a sync function). So, the plugin would be running on multiple different loops.

Which means an app thread waiting on the event could easily be in a different thread than the loop where the plugin is setting the event. Especially when you add in that apps can make their own threads.

Odianosen25:
Hello @zanfur 
> the app gets its own thread and all its calls to the plugin (via the api) are run in that thread's loop (or the executor if it's a sync function). So, the plugin would be running on multiple different loops

Oh now I see where the confusion is, and actually this is not the case. Even if its running from the app's thread, we do pass the function to be ran within AD's core event loop (we working on creating a different loop for apps, but that's another story). 

So don't be worried about the fact the function is called from the app's thread, as AD still executes service calls within its main loop.

zanfur:
Oh great! That makes things a bit easier. I just went and looked through
all of the app thread/worker selection and also realized that you only have
the single loop.  Which seems a really bad thing, because an app can make
an async function that just takes way too long and blocks all other apps
(and the core appdaemon), but if you say that's a problem for another time
then a single loop makes it a little less ugly to reuse the websocket.
Though it'll be harder when you actually move to multiple loops...

zanfur:
Technically the app can still make its own thread and loop but then the
plugin will complain when it tries to make the ws request in the executor,
because the executor loop doesn't match the future. Ask me how I know...

acockburn:
Did this PR die?

acockburn:
> Could we open this PR again? Seems like it's almost there and I'd be interested in seeing #505 resolved.

There was still a lot of work to do including the actual API, and docs. I am happy to revisit this if Zanfur (or anyone else) wants to complete this, but I am not able to accept partial PRs like this and do the rest of the work myself, I just don't have the bandwidth.

PR #1588, opened 2022-10-28
fix(app_managment): duplicated app service

chaoslion:
will working with the app reload service i found a duplicated register_service call.


acockburn:
Doesn't;t look like a dupe to me - one was restart, the other was reload.

chaoslion:
> Doesn't;t look like a dupe to me - one was restart, the other was reload.

foo

acockburn:
Got it - thanks!

PR #1606, opened 2022-11-22
Support entity.state in async contexts

### LINKED ISSUE #1605: entity.state not usable in async functions
Using `entity.state` inside an async function fails. This can be reproduced with a trivial app:

```python
import hassapi as hass

class MyApp(hass.Hass):
    async def initialize(self):
        entity = self.get_entity("zone.home")
        s = entity.state
```
This fails with a cryptic message (I guess it has to do with the internal use of `get_state` which is `async`):
```
2022-11-20 14:29:51.061465 WARNING myapp: Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/appdaemon/app_management.py", line 163, in initialize_app
    await init()
  File "/config/appdaemon/apps/myapp.py", line 6, in initialize
    s = entity.state
  File "/usr/lib/python3.10/site-packages/appdaemon/entity.py", line 532, in state
    return self.states.state
  File "/usr/lib/python3.10/site-packages/appdaemon/entity.py", line 19, in __get__
    stateattrs = utils.EntityStateAttrs(instance.get_state(attribute="all", copy=False, default={}))
  File "/usr/lib/python3.10/site-packages/appdaemon/utils.py", line 185, in __init__
    self.__dict__ = AttrDict.from_nested_dict(dict)
TypeError: __dict__ must be set to a dictionary, not a '_asyncio.Task'
```

Since the state can be fetched synchronously, and `entity.state` is much more readable than `await entity.get_state()`, it would be nice to support it. (If supporting it is not possible, it should at least be documented that `entity.state` is not available in async functions).

chatziko:
This tiny change allows using `entity.state` in async functions (which is much nicer than writing `await entity.get_state()`). The problem is that `entity.state` is a sync operation, so we need to fetch the state directly instead of via `entity.get_state` (which is an async operation).

Maybe there's a better way to do it, but I thought a PR is the best way to start a discussion :smile: 

Closes #1605

### DIFF
--- a/appdaemon/entity.py
+++ b/appdaemon/entity.py
@@ -16,7 +16,9 @@ class EntityAttrs:
         pass
 
     def __get__(self, instance, owner):
-        stateattrs = utils.EntityStateAttrs(instance.get_state(attribute="all", copy=False, default={}))
+        # Get the state directly (not via get_state) to be able to use entity.states in both sync/async contexts
+
+        stateattrs = utils.EntityStateAttrs(instance.AD.state.state[instance.namespace][instance.entity_id])
         return stateattrs
 
 


Odianosen25:
Thanks for this @chatziko, and really what you have suggested did really give us a potential way for fixing the issue. I think @acockburn will handle it and look into the PR

Thanks again

chatziko:
Thanks for looking into it! I've actually been testing this change in all my apps for a couple of months now, replacing all occurences of `await entity.get_state()` with `entity.state`. It seems to work fine, haven't noticed any side-effects.

Odianosen25:
Hello @chatziko,

Apologies for the late response, I missed the notification. The concept in principle is good, and in my view will work so no issues in that. It’s just the style we need to look into.

Like instead of accessing the AD object in the class, we considering creating a helper function in the state.py module which will essentially allow access to the state data similar to how you have done it. Then via that function you can then do what you have done.

On phone now so can’t write good code, but first to create a function “get_entity_state_simple”, which will return the state from the “self.state” object. Then use that in your code if that makes sense?

acockburn:
OK, I added this function along the lines that @odianosen suggested, it will be in the next AD release later today - thanks for the inspiration on how to fix this!

PR #1630, opened 2023-01-22
fix some details in appguide

moltob:
First of all: Thanks a lot for this doc page, it is a true jewel and exactly what I was missing in the jungle of docs and tutorials when approaching AD!

While reading and playing I found two leftovers from earlier AD versions I fixed.

### DIFF
--- a/docs/APPGUIDE.rst
+++ b/docs/APPGUIDE.rst
@@ -1384,7 +1384,7 @@ exactly like any other HA state, e.g., to the rest of AppDaemon, and the
 dashboard it looks like an entity. This means that you can listen for
 state changes in other apps and also publish arbitrary state to the
 dashboard via the use of specific entity IDs. To publish state, you will use
-``set_app_state()``. State can be retrieved and listened for with the
+``set_state()``. State can be retrieved and listened for with the
 usual AppDaemon calls.
 
 The Scheduler


acockburn:
Hi there - we really appreciate your efforts here, but unfortunately the corrections are not themselves correct. The import statement as shown is correct (AppDaemon adds the paths for the plugins to its module search path on startup), and `set_app_state()` is not a typo, it really is it's own API call.

If you disagree, and are having issues, by all mean post your errors here and we will get to the bottom of it.

moltob:
Thanks for your consideration. Let me try to explain how I got there.

``set_app_state`` was not offered by my IDE's code completion, so I started digging. If I search this repo, I only find (I assumed outdated) examples and this deprecation note for 4.0.0.b (it was "readded" in 3.0.0 _before_):

https://github.com/AppDaemon/appdaemon/blob/7314cb1591af4b3416584c6c0742604296145742/docs/HISTORY.rst?plain=1#L563

Regarding the import: You're correct. However, I do use full import paths for IDE code completion and seem to remember the original helloworld example in the HA addon was also using a full import path. So I got the impression this was changed at some point. 

I can update the PR to revert the imports if you confirm my first finding. Otherwise, I'd be happy to hear what I am missing and then please feel free to close the PR.

acockburn:
OK, good, you are right about set_app_state(), I remember now, it's been a while - thanks for catching that.

The imports are behaving as intended, it was a design decision to simply the sandbox and avoid arcane and nonsensical import statements. Of course there is no harm in doing it the way you suggest, it just wasn't what I was going for.

So, yep, go ahead and revert the imports and leave the fix for set_app_state() - thanks!

moltob:
Commit with imports was dropped.

acockburn:
Thanks

PR #1655, opened 2023-02-27
Dockerfile improvements

NigelRook:


Dockerfile
```diff
@@ -15,19 +26,18 @@ EXPOSE 5050
...
+# Copy compiled deps from builder image
+COPY --from=builder /usr/lib/python3.10/site-packages/ /usr/lib/python3.10/site-packages/
 
-# Install additional packages
-RUN apk add --no-cache curl
+# Copy appdaemon into image
+COPY . .
 
 # Start script
 RUN chmod +x /usr/src/app/dockerStart.sh
```
mion00, line 43: I would skip this step by simply making the file executable directly in the repo

Dockerfile
```diff
@@ -15,19 +26,18 @@ EXPOSE 5050
...
-# Install dependencies
-RUN apk add --no-cache build-base gcc libffi-dev openssl-dev musl-dev cargo rust curl \
-    && pip install --no-cache-dir .
+# Copy compiled deps from builder image
+COPY --from=builder /usr/lib/python3.10/site-packages/ /usr/lib/python3.10/site-packages/
 
-# Install additional packages
-RUN apk add --no-cache curl
+# Copy appdaemon into image
+COPY . .
```
mion00, line 40: We should copy only the files related to the application, instead of the whole repository. That way we have a better chance of re-using the Docker layer cache, without introducing unrelated changed files in the image.

Dockerfile
```diff
@@ -1,4 +1,15 @@
-ARG IMAGE=python:3.9-alpine
+ARG IMAGE=alpine:3.17
+FROM ${IMAGE} as builder
+
+WORKDIR /build
+
+# Install dependencies
+RUN apk add --no-cache git python3 python3-dev py3-pip py3-wheel build-base gcc libffi-dev openssl-dev musl-dev cargo
```
mion00, line 7: I suggest using the `RUN --mount type=cache` command to optimize the apk cache, as documented here:
https://docs.docker.com/build/cache/#keep-layers-small

Dockerfile
```diff
@@ -1,4 +1,15 @@
...
+
+# Fetch requirements
+COPY requirements.txt .
+RUN pip install -r requirements.txt
```
mion00, line 11: Since we have moved from a `requirements.txt` to a `pyproject.toml` to describe the dependencies, these step no longer work. 
I propose #1669 as an alternative.

acockburn:
Thanks both of you - it's about time Docker got a little love. We will optimize the image per Nigel's original intentions and add in Carlo's fixes.

Issue #1652, opened 2023-02-27
Suggestion: Improve CI pipeline


mion00:
## Introduction
Thanks for this amazing project!
I have been using it for quite some time to easily integrate an external device (a TP-link router) in Home Assistant.
After seeing the reported issues in the Discord chat about the docker builds failing, I would like to chime in and give an helping hand, having had some previous experience with CI systems (Gitlab CI and Github Actions) and Docker containers.
After a quick glance at the project, I have some questions regarding its development cycle:

## Continuous integration platform
It appears that multiple CI platforms are used:
- [semaphore ci](https://github.com/AppDaemon/appdaemon/blob/8de3a24ce447f954c8c7fe58d80eb6257c318508/.semaphore/semaphore.yml)
- [azure pipelines](https://github.com/AppDaemon/appdaemon/blob/8de3a24ce447f954c8c7fe58d80eb6257c318508/azure-pipelines.yml)
-  __Github Actions__: there has been a previous attempt in #1022 to introduce it, but at the moment in the `dev` branch the configuration has been removed, as seen from this commit: 363f1651fe5bbcb43814b712fa52be211946326f

### Question
Is there a reason why different CI platforms are needed? Wouldn't it be better to use a single one to consolidate the efforts and provide a single interface to see the latest build status?
Since the project is already using Github as its hosting platform, I am inclined to proposed __Github Actions__ as the CI provider, since it intuitively integrates with the rest of the ecosystem (github issues, pull request, releases, docker registry, etc..).

## Proposed changes
- [x] Open a discussion to decide if multiple CI platforms are indeed needed and preferable. In case a single one is preferred:
  - [x] Consolidate the workflow on a single platform to do automated testing and deployment
- [x] Define the automated steps that should happen in the CI pipeline (e.g.: should we run tests for every git push? create docker images on git tags?)
- [ ] Integrate the changes proposed in #1646 in a CI configuration (automatically generate release notes and release artifacts as part of the Github project)

## Implementation plan
General pipeline configuration:
- [x] Build dev as commits are pushed
- [x] Build releases whenever a TAG is created

The following is an outline of the ideal pipeline, summarized from the comments below:
- [x] Build and deploy docker multi arch images
- [x] Documentation (built using readthedocs)
- [x] Update dependencies (currently `dependabot` is used)
- [ ] Automatic release notes

acockburn:
Hi Carlo - and thanks for reaching out!

[...]

PR #1656, opened 2023-02-28
CI: basic GitHub actions

mion00:
This is a building block for #1652. It provides a simple setup for linting the Python code and running unit tests on every Git push and PR.

## Description
- Add Github Actions workflow that gets triggered on every Git push (inspired by [official Github examples](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python)):
  - the pipeline run against multiple python versions: "3.7", "3.8", "3.9", "3.10" (as does the current `azure-pipeline` config)
  - install dependencies from `requirements.txt` and cache them for faster builds
  - lint the code using `flake8`
  - run unit tests using `pytest`
- Add `tests/test_main.py` with sample unit test written for `pytest` framework (otherwise it errors out with `No tests found`)
- Add `pyproject.toml` to configure pytest (as from their [best practises docs](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html)).
- Move blake and flake8 config to `pyproject.toml` and `.flake8` files for better readability. This way it is possible to run both tools locally with the same config that is used in the CI pipeline (useful for troubleshooting possible config or linting mistakes).

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,9 @@
+[flake8]
+# Increase from the default 79
+max-line-length = 500
+
+extend-ignore = E203, E266
+# E203 whitespace before ‘,’, ‘;’, or ‘:’
+# E266 too many leading ‘#’ for block comment
+
+max-complexity=18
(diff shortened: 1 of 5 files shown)


acockburn:
Looks good to me after review. When I merge this should I remove the Azure integration since we don't need both?

mion00:
Yeah I suppose since we are migrating to Github Action it is ok to disable Azure pipelines, otherwise we have both running at the same time.

acockburn:
OK, I removed it in the dev branch – lets see how it goes 😊

Issue #1646, opened 2023-02-09
State of the Project: Releases for the small things


markusressel:
Hi there,

first of all, this is a really awesome project. It keeps my wife and me very happy in our little apartment home.

I know that maintaining projects like this is really hard. I try to do it myself and it just keeps getting harder every year. Splitting personal free time between projects like this, family, friends and hobbies is always a tough choice, and money (github f.ex. sponsoring) doesn't make it any easier.

Having said that, I would really like to see an update for appdaemon, since the last release is over a year old by now. Some small release for some basic dependency updates and python version migrations would already be great, just to see that the project isn't dead. I (and probably others as well) don't really need a ton of new features, just stability improvements, bugfixes and the like. Having to abandon appdaemon simply because it is getting too old would be a real shame. Afaik there isn't really an alternatives to it, and migrating to Home Assistant YAML files or the GUI... well it would suck. :sweat_smile: 

I have seen that there were some commits during the past year, so I assume its not dead, just on a very tight time-budget. I can understand that, and I would also like to help out with some small things if possible and/or necessary. I have to admit though, that I already am struggling to keep my projects alive and well too, so don't expect any wonders :smile: . While I know its not easy to find honest and hard working people to collaborate on a project like this, it could really help splitting the load, like it does for Home Assistant itself.

Thx for reading :heart:

acockburn:
Thanks for your concern – I can assure you that AD is alive and well and not going anywhere, not least because I recently moved into a new house and I am automating like crazy. Time is always an issue, I have a much more demanding job than when I started AD, and I’m spending a lot of time actually using it instead of developing it. Having said that, I am working towards a release soon – merging PRs for updates, I added a fix to the scheduler last weekend, so watch this space. In the meantime, we have a small but reasonably active Discord server here

https://discord.gg/q3P3qVkE

Drop in for a chat and to share experiences, there are usually few people around.

markusressel:
Well, you still using your own project is great to hear! 😄 
I will make sure to check out the discord.

I want to emphasize tough that my main point is still the release cycle. I think if making releases was easy (not sure how easy it is currently), and release notes would actually be part of the github release page (there are bots for this), or at least a link to the documentation, I think it would make the project appear much more alive and probably attract a lot of users that might be scared away by old releases and "unpolished" update notes. Even I was starting to get worried even though the currently latest release works pretty darn well on a daily basis.

Thx for the quick response! ❤️

acockburn:
[...]

PR #1661, opened 2023-02-28
Improve packaging of application

### LINKED ISSUE #1657: Improve packaging system
The `requirements.txt` file is used in a bit confusing way. It it usual practice for this file to contain all the dependencies present in the project virtualenv, created using `pip freeze > requirements.txt`, so that the virtualenv can be easily re-created by any developer locally using `pip install -r requirements.txt.`
But this file is also used as the `install_requires` inside `setup.py`, creating a bit of a conflict.
If one installs the appdaemon locally with `pip install .`, then run `pip freeze > requirements.txt`, the `requirements.txt` file is modified, since all the transitive dependencies are also loaded, as is expected. So the question is: shouldn't the declaration of the project dependencies be moved in a more appropriate place, separate from `requirements.txt`?

## Proposed solution
I propose to more crealry declare the project dependencies, by upgrading the way we use the Python packaging system.
At the moment the package is created from the `setup.py`, read by `setuptools` when building the package.
The way to package applications has changed a lot over the years. Nowadays if one were to follow the official [Python packaging guide](https://packaging.python.org/en/latest/tutorials/packaging-projects/), the recommended setup is to use a `pyproject.toml` to describe all the metadata of the project, including the build tool and the project dependencies.
I suggest we move from the current `setup.py` to a more modern and "standard" solution, by migrating to a `pyproject.toml` file to contain all the project dependencies.

mion00:
Fixes #1657

## Description
- Describe project metadata in `pyproject.toml`
- Remove `requirements.txt`: dependencies are now part of the `pyproject.toml` metadata
- Remove `setup.py`: the recommended way to package a Python application is via a `build-system` configured as part of the  `pyproject.toml`. 
- Define development dependencies under optional `dev` group
  - Can be installed by `pip install .[dev]` (the `[ ]` part specifies a package [extras](https://packaging.python.org/en/latest/tutorials/installing-packages/#installing-extras), a variant of the package with additional dependencies. In this case, the development dependencies)
- Update development documentation with instructions on how to setup a local development environment and publish a Python package
- Link development documentation from README for ease of access
- Update Github pipeline to use the new installation method instead of `requirements.txt`


acockburn:
This looks great - very pleased that you updated the docs too, I compliment you on your thoroughness! I'll merge this shortly, then I will need to figure out how to build for PyPi but that shouldn't be a big deal, and is not urgent.

mion00:
> This looks great - very pleased that you updated the docs too, I compliment you on your thoroughness! I'll merge this shortly, then I will need to figure out how to build for PyPi but that shouldn't be a big deal, and is not urgent.

I added a section in the docs explaining just that! :wink: 
The next logical step is automating the PyPi build as part of the CI, so event that part is out can be out the equation.
Since this might seems a big change in the build process, let me know if anything breaks, and if I can be of help.

acockburn:
Haha, I missed that - it's not always easy to see everything from the deltas that GitHub serves up. But it's merged so I'll go read it now - thanks.

Automating the PyPi build would be cool, didn't know that was possible, but of course it is :)

I usually only update PyPi for actual releases, although I occasionally put out a beta. So, this isn't a huge change, it just cuts out a manual step once per release. I always tag releases, so the rule could be based on that. I have also just decided to start properly using GitHub's release feature - I know that too is based on tags, does that make any difference?

mion00:
I think we can approach the situation from both angles: Github Actions supports triggering from both tags and release, so we can decide what is the preferred way:
- manually create a git tag, push the tag -> automatic build and release on PyPi -> automatic creation of a `release` in Github

- manually create a git tag, push the tag, then
   - manually create a Github release ->  triggers a build and release on PyPi

Of course there could be subtle differences in one approach over the other, I am just brainstorming some possible ideas. Depends also on the solution proposed by https://github.com/AppDaemon/appdaemon/issues/1646#issuecomment-1445440644 about automatic release notes, that _Release Drafter_ is something new also for me.

acockburn:
Interesting. I think my preferred workflow would be to create the TAG manually, Create the release from the tag then have PyPi generation just happen.

Reason is, I will be wanting to add comments to the release, and also we manually keep a history file up to date which is a little more informative than just listing the PRs, so I would like to manually include that as part of the release along with my comments, then finally add all the PRs etc. at the bottom. That would be ideal for me I think.

PR #1666, opened 2023-02-28
CI: automatic build

mion00:
Automatically build the Docker container and the Python package.

## Description
- On pull requests against `dev` -> build Python package and the Docker image, but don't publish them.
- When there is a push in the `dev` branch -> build Docker image with tag `dev`, deploy to Docker Hub
- When there is a git tag, build Docker image with tag taken from version + deploy to Docker Hub and PyPi.

## TODO
How to handle version numbering for the Python package? Should we use the Git tag to automatically infer the package version?
That way we don't have to manually update the `version.py` file every time we have a release, risking a mismatch between the tag and the version written inside if one forgets to keep them in sync.

## Requirements
Setup [Github secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets#about-encrypted-secrets) inside the repository settings with the credentials to authenticate against PyPI and Docker Hub
Secrets needed:
- PYPI_API_TOKEN [instructions on how to create API token](https://pypi.org/help/#apitoken)
- DOCKERHUB_USERNAME [instruction on how to create access token](https://pypi.org/help/#apitoken)
- DOCKERHUB_PASSWORD

### DIFF
--- a/.github/workflows/build-deploy.yml
+++ b/.github/workflows/build-deploy.yml
@@ -33,7 +33,7 @@ jobs:
         uses: actions/checkout@v3
       # https://github.com/actions/setup-python
       - name: Set up Python
-        uses: actions/setup-python@v3
+        uses: actions/setup-python@v4
         with:
           python-version: "3.10"
           cache: 'pip' # caching pip dependencies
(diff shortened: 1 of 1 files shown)


acockburn:
Excellent – I will merge shortly.

Regarding the version, I have been updating manually up to now, but I would be happy if everything could be versioned from the TAG when I create it. I guess the dev versions might be a problem though.

PR #1669, opened 2023-03-01
CI: Optimize docker build

mion00:
This PR is built on top of #1655, augmenting it with the new project setup (`pyproject.toml` instead of `requirements.txt`) and CI pipeline.

## Description 
- Always build the Docker image and Python package (pushing only with tag or `dev` branch) as part of the Github pipeline.
This way we can easily catch any problem in the build process, without having to first merge the changes inside `dev`.
- Use Python package built during the `build_package` job for building the Docker image. This way we avoid having to copy the entire project source inside the image, reducing the image size.
At the same time this process "mimics" the step an end-user of the package does (a `pip install`), so we can verify that the package can indeed install correctly.
- Optimize the docker build cache by using `RUN --mount=type=cache` to speed up the process between subsequent builds.
- Small fix to the `pyproject.toml`: exclude the static assets inside `docs.` folder, otherwise they get included as part of the Python package, unnecessarily bloating it

### DIFF
--- a/.github/workflows/build-deploy.yml
+++ b/.github/workflows/build-deploy.yml
@@ -21,6 +21,7 @@ env:
   IMAGE_NAME: acockburn/appdaemon
 
 jobs:
+  # The Python package is required in the subsequent Docker image build
   build_package:
     name: Python package
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


acockburn:
The build failed with the new docker file:

https://github.com/AppDaemon/appdaemon/actions/runs/4303953648/jobs/7504338740

mion00:
Yeah the changes from #1655 where not exactly up-to-date with the state of the projects, for instance the use of `requirements.txt` breaks the build.
I am working on a fix in this PR.

PR #1671, opened 2023-03-01
CI: Update dependabot config

mion00:
## Description
- remove `ignored` rules from pip updates, since we are using more recent versions of those dependencies, so telling the bot to ignore those old versions has no sense
- monitor Docker updates
- monitor Github Actions update

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,6 @@
 version: 2
 updates:
+# Enable version updates for pip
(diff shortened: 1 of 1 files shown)


acockburn:
This one seems to be failing CI, but all of the actions succeeded ...

mion00:
The last action has failed (you need to scroll a bit inside the window, it's not immediately visible). It is failing due to the latest arm changes.

PR #1673, opened 2023-03-02
Fix `arm/v6,v7` builds, pin dependency versions in `requirements.txt`

mion00:
Fix 36dc3df with the addition of `arm/v6` and `arm/v7` Docker builds.
This is also a follow-up from: #1661. I am sorry if this seems like a backtrack on my past decisions, but I refreshed my knowledge on current Python packaging best practices in the last days (something that continues to be quite messy with the number of tools available), and on how to appropriately manage application dependencies.

## Description
- Re-introduce `requirements.txt` files **pinning** the specific version of dependencies we are using.
A nice explanation as to why `requirements.txt` file is recommended can be found [here](https://caremad.io/posts/2013/07/setup-vs-requirement/) (it was written when `setup.py` was the standard, nowadays we use `pyproject.toml` instead, but the key concept still remain the same).
Since AppDaemon is published as a Python package, but at the same time runs as a standalone application, the recommendation is to use a `requirements.txt` file to specify the exact version, while keeping versions specified in `pyproject.toml` as "abstract" dependencies.
  - This is also required to optimize the build of the Docker image, by caching the dependencies separately from the application code 
- Introduce `pip-compile` to **pin** the version of dependencies and automatically manage the `requirements.txt` files, each one appropriate for a specific environment (application runtime, local dev, doc building)
- Separate dependencies required to build documentation in a `doc` dependencies group
- Re-introduce the Docker multi-build stage previously defined in #1655 , to optimize the re-use of cached dependencies in subsequent Docker builds.
- Update development documentation explaining how the Docker build works, and detailing the dependency management of this project
- Use `pre-commit` hook from `pip-compile` to automatically keep the `requirements.txt` files in sync whenever there is a change to `pyproject.toml`

**NOTE**: the Docker build will be quite long the first time, since it has to compile from source the Python packages not available for `arm` architectures (`orsjson` and `uvloop`).
Thanks to Docker build cache however, subsequent builds should be blazing fast, re-using the already compiled packaged.


acockburn:
All good - I'll fix the docs later :)

One question - how/when does pip-compile get run, is this a manual step ahead of each release?

mion00:
The purpose of `pip-compile` is to "lock" the version of dependencies inside "requirements.txt".
I manually invoked it the first time to get these requirements file here. If I understood its usage correctly the only time we need to invoke it is when we want to update a dependency, using its [--upgrade or --upgrade-package flags](https://pip-tools.readthedocs.io/en/latest/#updating-requirements)

mion00:
There is also a `pre-commit` hook ready, in case it might come in handy (https://pip-tools.readthedocs.io/en/latest/#version-control-integration)

acockburn:
I merged these out of order and have a conflict - could you fix it, sorry. I don't want to break anything!

PR #1687, opened 2023-03-03
Enable use of alpine py3-* images in docker

NigelRook:
Fixed py3-* packages. Alpine puts python in /usr/lib/python3.10 whereas the python images put it in /usr/local/lib/python3.10, so added the /usr/lib paths to PYTHONPATH.

Also added some comments and substitutions to guard against future updates causing a version mismatch between the python image and alpine.

py3-* images install a second copy of python doing things this way. That might potentially be a good thing if it stops alpine overwriting specific versions of packages pinned by appdaemon.

Tested build for all 4 platforms, worked fine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,7 @@
-ARG BASE_IMAGE=python:3.10-alpine
+# When changing this, make sure that the python major and minor version matches that provided by alpine's python3
+# package (see https://pkgs.alpinelinux.org/packages), otherwise alpine py3-* packages won't work
+ARG PYTHON_RELEASE=3.10 ALPINE_VERSION=3.17
+ARG BASE_IMAGE=python:${PYTHON_RELEASE}-alpine${ALPINE_VERSION}
 # Image for building dependencies (on architectures that don't provide a ready-made Python wheel)
 FROM ${BASE_IMAGE} as builder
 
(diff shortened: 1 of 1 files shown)


acockburn:
@NigelRook I'm interested in moving to Python 3.11 for the docker image. It seems as simple as changing the version in the variable to 3.11 as there is a python alpine image for this, However, I am concerned by your comment:

```
# When changing this, make sure that the python major and minor version matches that provided by alpine's python3
# package (see https://pkgs.alpinelinux.org/packages), otherwise alpine py3-* packages won't work
```

From the link I am unable to figure out which version alpine's python provides (I think it's 3.10 in 3.17) if this is correct does that mean that installation of python 3.11, while it will seem to work, won't allow installation of APK packages?

For fun, I made the change and tried it locally and it worked just fine, but I don't install anything additional.

NigelRook:
Correct. Technically, the packages will install, but import <package> will fail.

Testing this shold look something like (snippet from head and untested):

```
docker exec -it <appdaemon container> /bin/sh

$> apk add py3-scikit-learn
$> python3

>>> import sklearn
```

If that works you're good

Alpine edge is running python 3.11 if you're desperate for it, so you can try setting ALPINE_VERSION to 20230208 and see if that works. Obviously that's a pre-release distro with the usual caveats.

acockburn:
OK, that's very helpful - and thanks for the quick response. I'll stay away from edge and wait until 3.18 is out which will presumably have python 3.11, and I'll figure out a testing methodology along the lines you suggested.

PR #1695, opened 2023-03-07
ci(build-deploy.yml): correctly set Docker latest tag only on release versions

mion00:
The default configuration for the `docker/metadata-action@v4.3.0` is to tag with `latest` whenever there is a git tag, but this is not compatible with our use of `pre-release` tags (alpha, beta, etc..).
This versions should still be tagged, just not with `latest`.

## Description
- Tag with `latest` the git tags that do not have a "pre-release" component in the end (e.g. `3.0.0`)
- Avoid tagging with `latest` the git tag that have a "pre-release" component in the end (e.g. `3.0.0b1`)
- If no git tag, fallback to using branch or PR name as tag

### DIFF
--- a/.github/workflows/build-deploy.yml
+++ b/.github/workflows/build-deploy.yml
@@ -134,6 +134,21 @@ jobs:
         uses: docker/metadata-action@v4.3.0
         with:
           images: ${{ env.IMAGE_NAME }}
+          # Customize the generation of Docker `latest` tag
+          # Tag with `latest` the git tags that do not have a "pre-release" component in the end (e.g. `3.0.0`)
+          # Avoid tagging with `latest` the git tag that have a "pre-release" component in the end (e.g. `3.0.0b1`)
+          # If no git tag, fallback to branch or PR name
+          tags: |
+            # If the git tag follows PEP440 conventions, use it as the resulting docker tag (both releases and pre-releases)
+            type=pep440,pattern={{version}}
+
+            # If the git tag does NOT have a pre-release ending (e.g. `3.0.0`), it is a release version to be tagged as `latest`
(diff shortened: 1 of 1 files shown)


mion00:
May I suggest pushing a `beta` version out sooner or later? That way we could test the whole deploy pipeline (if everything goes to plan, we should avoid breaking existing users installations).
With this fix we won't touch the `latest` tag in Docker using a `pre-release` tag, so existing users will not pull down the beta by accident. Also `pip` installation automatically skip `beta` versions (if I read the docs correctly).

acockburn:
Definitely - it should be as simple as tagging the current master branch to make a release, it will have zero changes and wouldn't;t contain any dangerous breaking code but would exercise the release process. Afterwards it will be easy to clean up the releases and any artifacts that make it out to PyPi or Docker by mistake. Sound like a plan?

mion00:
> Definitely - it should be as simple as tagging the current master branch to make a release, it will have zero changes and wouldn't;t contain any dangerous breaking code but would exercise the release process. Afterwards it will be easy to clean up the releases and any artifacts that make it out to PyPi or Docker by mistake. Sound like a plan?

If you tag the current `master` I think nothing will happen. It doesn't contain the github workflow files, since they are all in the `dev` branch.

acockburn:
You make a good point ;)

ok well I’ll do a release in a day or two of whatever is in dev and call it a beta

acockburn:
OK, I got around to trying out a beta release and noticed a couple of things:



  1.  The package was pushed to PyPi but the version label applied was “4.3.0” instead of “4.3.0b1”, I tagged it that way in Github, and marked it as a prerelease
  2.  The docker build didn’t seem to be pushed to Docker, although the dev builds have happily been making it over there

Any thoughts?

Andrew

PR #1749, opened 2023-04-12
remove extra "_" in input_datetime.yml static CSS

D3SL:
```container_style: $input_datetime_container_style``` contained an extra underscore which prevented CSS styling users tried to pass from actually being used.


acockburn:
Thanks!

PR #1750, opened 2023-04-12
Fix for template/render service

Justihar:
Resolves an issue that uses in invalid URL while processing template/render service in the HA plugin


rcmurphy:
Any chance you could cut 4.4.2 with this fix?

acockburn:
Yes, I'll take a look later today.

acockburn:
OK, this is in 4.4.2 which is just about to be released.

PR #1753, opened 2023-04-13
Fix for events listening to event

Odianosen25:
This fixes an issue whereby due to the new dictionary unpacking, it broke the plugins ability to listen to event code.

So this fixes it, and makes it that the default behaviour for plugins is to use the dict unpacking and the user has to forcefully use otherwise.


acockburn:
Why are we switching the default for dict unpacking? I don't want to do this for a while.

acockburn:
Not sure what you mean about the default unpacking behavior here?

PR #1838, opened 2023-07-08
Babel package

zanaca:
Adds Babel as requirement to enable i18n  off apps, as seen in [joBr99/nspanel-lovelace-ui](https://github.com/joBr99/nspanel-lovelace-ui/blob/main/docs/prepare_ha.md)

I have been adding it through an extra Dockerfile import for a while and it is working fine.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.woff2 binary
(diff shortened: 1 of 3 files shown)


acockburn:
Hi there - not sure I understand what this adds. I agree internationalization is desirable and maybe one day AD will have it, but in the meantime, this just adds an unused requirement to the AD build, which will be installed and unused in VENVs and the Docker build. There exist manual ways to do this for folks who want to use this.

Perhaps you can explain better what problem this solves?

Issue #1857, opened 2023-11-14
Icon widget extension: service calls - ignored for default state


lancasterJ:
### What happened?

The icon widget supports post_service_active for calling a Homeassistant service when the icon is clicked.  This does not work when sensor state is not listed in the state map and the "default" rule is matched.

I believe the cause of the issue is line 89 in baseicon.js ([https://github.com/AppDaemon/appdaemon/blob/02a45c36c0e4d22ae14e630e7e747406fad4b753/appdaemon/widgets/baseicon/baseicon.js#L89-L90]) 

It is currently 
`set_service_call(self, self.parameters.default);`

and should be

`set_service_call(self, self.parameters.icons.default);`

### Version

4.4.2

### Installation type

Home Assistant add-on

### Relevant code in the app or config file that caused the issue

```sh
maintenance_alert:
    widget_type: icon
    entity: sensor.maintenance_alert
    title: Maintenance
    title2: summary
    state_text: 1
    icons:
      "No alert":
        icon: mdi-progress-clock
        post_service_active:
          service: homeassistant/turn_on
          entity_id: script.ad_load_dashboard_maintenance_pi_toilet         
      "default":
        icon: mdi-alert
        post_service_active:
          service: homeassistant/turn_on
          entity_id: script.ad_load_dashboard_maintenance_pi_toilet
```

acockburn:
fixed in dev

PR #1913, opened 2024-01-26
Fix for issue #1857

lancasterJ:
Updated baseicon.js to pass correct second argument to set_service_call in function set_view when the default icon is used.


acockburn:
thanks!

lancasterJ:
Hi Andrew ( @acockburn )
Do you have a target date for when a release that includes this fix will be made?
Regards
John

acockburn:
We are due a release but I have had a lot of business travel for my job recently. This will definitely be in the next release, and I will hopefully get to it soon, I’m sorry I can’t be any more accurate than that!

lancasterJ:
> We are due a release but I have had a lot of business travel for my job recently. This will definitely be in the next release, and I will hopefully get to it soon, I’m sorry I can’t be any more accurate than that!

I fully understand. Thanks for the update.

PR #2114, opened 2024-10-04
Fix type hint

poggenpower:
datetime.datetime.utcnow().timestamp() returns a float.  The given example already shows this correct.


acockburn:
Thanks!

PR #2161, opened 2024-11-24
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

Redoing my HASS installation from scratch and going through components, contributing where I can ;-)

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,25 @@
+# Codespell configuration is within pyproject.toml
+---
+name: Codespell
+
+on:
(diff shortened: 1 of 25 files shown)


acockburn:
This is really neat thanks! I'll merge this soon, and I have started using this in other projects :)

acockburn:
Thanks, but it's already in and released :)

PR #2178, opened 2024-12-14
Improve sync_wrapper's definition so that we get better typing

Ten0:
Before, calling functions such as `self.get_state` or `self.turn_on` with anything else but the required type (in these two cases, str) would not lead to a type error, but rather to a runtime error long down the chain.

This now properly fails to typecheck.

```
Argument of type "whatever" cannot be assigned to parameter "entity_id" of type "str"
```

### DIFF
--- a/appdaemon/entity.py
+++ b/appdaemon/entity.py
@@ -103,7 +103,7 @@ class Entity:
 
     @utils.sync_wrapper
     async def get_state(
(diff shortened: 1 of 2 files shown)


Ten0:
Ah yeah now it's called sync_decorator, and the issue is the other way around... Before we were losing all the typing, whereas now the typer keeps thinking that the function returns a coroutine, although it doesn't if running in a non-asynchronous context:

image

PR #2215, opened 2025-04-14
fix for client_topics

jsl12:
The string `'NONE'` wasn't getting appropriately translated into an empty list


jsl12:
```yaml
appdaemon:
  ...
  plugins:
    ...
    mqtt:
      ...
      client_topics: NONE
```

Will now result in an empty list of `client_topics`, which means no subscriptions get made

PR #2301, opened 2025-05-29
added a pip upgrade line to the dockerfile

### LINKED ISSUE #2288: A new release of pip is available: 25.0.1 -> 25.1.1
### Is there an existing feature request for this?

- [x] I have searched the existing issues

### Your feature request

When starting AD, you get the message logged:
`[notice] A new release of pip is available: 25.0.1 -> 25.1.1`
Maybe to update pip for a next AD (patch) release.

jsl12:
Fixes #2288


chatziko:
This does not seem like a proper solution to me:
1. By the time the container runs there might be an even newer version of pip, so the problem is not solved
2. It causes a larger docker image.

Better to pass `--disable-pip-version-check` to pip install, this version check makes little sense for containers.

hacscred:
The right solution is to drop Docker support and just use Nix (from which OCI containers can be built with existing generic tooling).

hacscred:
AFAIK, docker builds OCI images: https://nix.dev/tutorials/nixos/building-and-running-docker-images.html and even if it doesn't OCI images are still somewhat "advanced" for most people I the sense that those people looking for a Docker image don't care how they are created. 

For any project with dependencies packaged in nixpkgs already, it should be preferable to tracking any random distribution, because in particular you can pin versions to make sure everyone runs the same thing and you only need to support _one_ combination of software, instead of billions.

jsl12:
Like I mentioned, I don't really have any experience with building Docker images from nix, but this is my general understanding of the situation:

- Docker was initially based on LXC containers and eventually morphed into what it is today. They essentially created the OCI standard, so their images are of course compliant
- The OCI standard is just about the resultant container, so they can be built in different ways.
- Nix is able to build the containers with layers side-by-side instead of top-down because of the isolation of the nix store
- When a dependency changes, only the layer for that dependency changes

hacscred:
Yes, all of that is correct. I have used this technology in arguably a critical production environment; it works.

markusressel:
Seems a bit excessive to me to rewrite the whole container toolchain because of a warning that doesn't really affect anything. Wouldn't it be good enough to add `--disable-pip-version-check` for the time beeing and create a separate issue/PR about the toolchain stuff?

hacscred:
I think this is not a corporate project, but if you want to feel corporate you can do that. 

I think what I described is just a perfect solution, but that doesn't mean awful hacks can't be implemented until better solutions arrive. 

Also, this is just my opinion. I have no authority here.

jsl12:
To be clear, if we were to start a nix-built OCI image, it wouldn't be to address this pip issue, although it would. It would be because it'a a good idea generally. It would allow anyone with nix installed (and flaked enabled) to set up a complete dev environment with variables, shell aliases, build tools, everything, all installed and configured to every minute detail, with a _**single command**_. You wouldn't even have to clone manually
 
Imagine...
```shell
nix develop github:appdaemon/appdaemon
```

For example, this is what my devenv setup looks like:

- Neither uv or python are even installed to begin with
- `ads` is my alias for "AppDaemon shell", which just activates the shell provided by the flake
- There's a nice printout of what URL , branch, commit hash, and versions you're on when the shell activates
- Once in the shell, you can use `uv` and `python` commands normally

image

PR #2307, opened 2025-05-29
Fix wait_state (broken in 4.5.0)

chatziko:
A `finally` clause was added to `wait_state` in 4.5.0 to remove the event from `_async_events` in case `entity_state_changed` is never called. However, this creashes in the normal case then  `entity_state_changed` __does__ get called, cause `async_events.pop`  does not find the key and fails.

This tiny PR fixes this bug.


acockburn:
Thanks!

Issue #2282, opened 2025-05-27
web_protocol.py issue with prepare_meth


Baggypants:
### What happened?

Not quite sure what to make of this

```
2025-05-27 23:24:08.501071 INFO AppDaemon: Loading Plugin HASS using class HassPlugin from module appdaemon.plugins.hass.hassplugin
2025-05-27 23:24:08.501286 INFO HASS: HASS Plugin initialization complete
2025-05-27 23:24:08.505067 INFO AppDaemon: Initializing HTTP
2025-05-27 23:24:08.505310 INFO AppDaemon: Using 'ws' for event stream
2025-05-27 23:24:08.505649 INFO AppDaemon: HTTP Listening on port 5050
2025-05-27 23:24:08.507342 INFO AppDaemon: Starting API
2025-05-27 23:24:08.508444 INFO AppDaemon: Starting Admin Interface
2025-05-27 23:24:08.508623 INFO AppDaemon: Starting Dashboards
2025-05-27 23:24:08.514460 INFO AppDaemon: Starting Apps with 1 workers and 1 pins
2025-05-27 23:24:08.516643 INFO AppDaemon: Running on port 5050
2025-05-27 23:24:08.516866 INFO AppDaemon: Waiting for plugins to be ready
2025-05-27 23:24:08.542092 INFO HASS: Connected to Home Assistant 2025.5.1 with aiohttp websocket
2025-05-27 23:24:08.543226 INFO HASS: Authenticated to Home Assistant 2025.5.1
2025-05-27 23:24:08.544378 INFO HASS: Waiting for Home Assistant to start
2025-05-27 23:24:08.550455 INFO AppDaemon: All plugins ready
2025-05-27 23:24:08.551440 INFO AppDaemon: Scheduler running in realtime
2025-05-27 23:24:08.570649 INFO AppDaemon: New app config: hello_world
2025-05-27 23:24:08.580832 INFO HASS: Completed initialization in 70.919ms
2025-05-27 23:24:08.583516 INFO AppDaemon: Starting apps: {'hello_world'}
2025-05-27 23:24:08.584991 INFO AppDaemon: Calling initialize() for hello_world
2025-05-27 23:24:08.634094 INFO hello_world: Hello from AppDaemon
2025-05-27 23:24:08.634736 INFO hello_world: You are now ready to run Apps!
2025-05-27 23:24:08.635507 INFO AppDaemon: App initialization complete
[...]

PR #2313, opened 2025-05-30
fix: add name property to RequestHandler

sholdee:
The exception handler fix for https://github.com/AppDaemon/appdaemon/issues/2282 got me further:

```js
Error in stream: Unknown error occurred, check AppDaemon logs: 'RequestHandler' object has no attribute 'name' {response_type: 'hello', response_id: '9cfb952e-33cc-4304-b620-4e1b2a359e44', response_success: false, response_error: "Unknown error occurred, check AppDaemon logs: 'RequestHandler' object has no attribute 'name'", request: {…}}
```

Adding this then fixed the issue, and data now loads in the admin dashboard.

### DIFF
--- a/appdaemon/stream/adstream.py
+++ b/appdaemon/stream/adstream.py
@@ -102,6 +102,11 @@ class RequestHandler:
         self.stream = self.adstream.stream_handler.makeStream(self.AD, request, on_message=self._on_message, on_disconnect=self._on_disconnect)
         #
 
+    @property
+    def name(self):
+        """Return client_name as name for compatibility"""
+        return self.client_name
+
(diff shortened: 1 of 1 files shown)


sholdee:
There are still issues actually. The apps, plugins, threads, and callbacks populate, but all the metrics are 0 values. Uptime is also showing a negative value `Uptime: -1 day, 23:53:30` and decrementing. If `http.password` is enabled, the password is never accepted and the prompt comes back immediately.

acockburn:
I think this should all be fixed in 4.5.5 - the negative uptime was a refactoring issue, and we added an alternate fix for the property issue.

PR #2347, opened 2025-06-03
Feature/add skin to navigate

### LINKED ISSUE #2345: Feature Request: Add "skin" Parameter to dash_navigate Function
### Is there an existing feature request for this?

- [x] I have searched the existing issues

### Background
I'm implementing a power-saving mode for my AppDaemon dashboard that would automatically switch to a minimalist "power saving" dashboard (essentially a black screen with a clock) when no one is present in the room, and return to the main dashboard when someone enters the room.

### Feature Request
Please add a `skin` parameter to the `dash_navigate` function that would allow changing the skin/theme when navigating between dashboards.

### Use Case Example
```python
# When room becomes vacant
def on_room_vacant(self, entity, attribute, old, new, kwargs):
    self.dash_navigate(self.args["power_saving_dashboard"], skin="dark")

# When someone enters the room
def on_room_occupied(self, entity, attribute, old, new, kwargs):
    self.dash_navigate(self.args["main_dashboard"], skin="default")
```

### Question to the Team
Is there perhaps a better approach to implement this power-saving functionality? I'm open to alternative solutions if the current dashboard framework already has capabilities that could address this use case without requiring this specific feature addition.

**Note**:  If you prefer I can open a PR for that as it does not seem to be a complex change.

Thank you for considering this request!

vitorrm:
# Pull Request Description for dash_navigate Skin Parameter Feature

## Title
Add skin parameter to dash_navigate function for dynamic theme switching

## Description

This PR adds a new optional `skin` parameter to the `dash_navigate` function, allowing users to change dashboard skins/themes during navigation. 

### What this PR accomplishes:

- Extends the `dash_navigate` function to accept and process a `skin` parameter
- Maintains backward compatibility with existing code (parameter is optional)
- Tiny refactor on the update_dash function for better handling of the query params
- I've tried to avoid any new features from JS as I think this project tries to keep compatibility with older browsers (my own case - IPad 3)

### Use case:
- Implementing power-saving modes for dashboards (switching to a dark theme when a room is vacant)

### Example usage:
```python
# Switch to a dashboard with a specific skin
self.dash_navigate("main_dashboard", skin="dark")

# Standard usage without skin parameter still works
self.dash_navigate("settings_dashboard")
```

### Testing:
I've tested this functionality and verified that the skin parameter correctly applies the specified theme during navigation while preserving all existing functionality.

Closes #2345

### DIFF
--- a/appdaemon/adapi.py
+++ b/appdaemon/adapi.py
@@ -3294,7 +3294,7 @@ class ADAPI:
     # Dashboard
     #
 
-    def dash_navigate(self, target: str, timeout: int = -1, ret: str | None = None, sticky: int = 0, deviceid: str | None = None, dashid: str | None = None) -> None:
+    def dash_navigate(self, target: str, timeout: int = -1, ret: str | None = None, sticky: int = 0, deviceid: str | None = None, dashid: str | None = None, skin: str | None = None) -> None:
         """Forces all connected Dashboards to navigate to a new URL.
 
         Args:
@@ -3315,6 +3315,7 @@ class ADAPI:
             dashid (str): If set, all devices currently on a dashboard which the title contains
                 the substring dashid will navigate. ex: if dashid is "kichen", it will match
                 devices which are on "kitchen lights", "kitchen sensors", "ipad - kitchen", etc.
+            skin (str): If set, the skin will change to the skin defined on the param.
 
         Returns:
             None.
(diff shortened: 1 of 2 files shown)


vitorrm:
That is fine for me, totally get it

acockburn:
OK, this will be in 4.5.11, thanks again!

Issue #2362, opened 2025-06-07
Registering a service in an AppDaemon app fails with: TypeError: Services.register_service() missing 1 required positional argument: 'callback'


mattbod:
### What is the issue?

Registering a service in an AppDaemon app fails with:
TypeError: Services.register_service() missing 1 required positional argument: 'callback'

This happens for any user code, even a minimal "Hello World" test with:
    self.register_service("hi", self.dummy)

Tested on Home Assistant OS 15.2, AppDaemon add-on version 0.17.4, AppDaemon core 4.5.8, Python 3.12.10. Reproducible after full add-on removal, config wipe, host reboot, and fresh install.

### Steps to reproduce

1. Clean install Home Assistant OS 15.2 and latest AppDaemon add-on.
2. Place this minimal app in `/config/apps/hello.py`:

```python
import appdaemon.plugins.hass.hassapi as hass
class Hello(hass.Hass):
    def initialize(self):
        self.log("Hello world!")
        self.register_service("hi", self.dummy)
    def dummy(self, namespace, service, data, kwargs):
        self.log("You called me!")

3. Observe error in logs:

TypeError: Services.register_service() missing 1 required positional argument: 'callback'


LOGS:

2025-06-07 18:19:34.587566 ERROR Error: ===========================================================================
2025-06-07 18:19:34.587813 INFO AppDaemon: App initialization complete
2025-06-07 18:19:34.790803 INFO AppDaemon: New client Admin Client connected
2025-06-07 18:19:46.690633 INFO AppDaemon: Modified Python files: 1
2025-06-07 18:19:46.691860 INFO AppDaemon: Modification affects apps {'remote_brush_light'}
2025-06-07 18:19:46.695630 INFO AppDaemon: Stopping apps: {'remote_brush_light'}
2025-06-07 18:19:46.696854 INFO AppDaemon: Starting apps: {'remote_brush_light'}
2025-06-07 18:19:46.703152 INFO AppDaemon: Calling initialize() for remote_brush_light
2025-06-07 18:19:46.708317 INFO remote_brush_light: RemoteBrushLight initializing���
2025-06-07 18:19:46.710406 WARNING AppDaemon: App 'remote_brush_light' failed to start
2025-06-07 18:19:46.711247 ERROR Error: =====  Failed to start 'remote_brush_light'  ==============================
2025-06-07 18:19:46.711810 ERROR Error: AppStartFailure: App 'remote_brush_light' failed to start
[...]

PR #2364, opened 2025-06-07
improved error text

jsl12:


jsl12:
This should help with situations like #2362

Issue #2188, opened 2024-12-31
run_at_sunset() with offset not handled correctly after a reload


ekutner:
### What happened?

When a task is scheduled with run_at_sunset() with an offset and the code is reloaded or the server is restarted in the time frame between sunset and sunset+offset the task is scheduled to run the next day instead of the current day.

Example:
Assume sunset is at 17:00 and there is an offset of 3600s, which means the task should run at 18:00.
Restart the appdaemon server any time between 17:00 and 18:00
The task should still be scheduled for the same day at 18:00 because that time didn't occur ye,t however, it is scheduled for the next day.

### Version

4.4.2

### Installation type

Home Assistant add-on

### Relevant code in the app or config file that caused the issue

```sh
self.run_at_sunset(self.do_ops, offset=3600)
```

ekutner:
I'm not sure why this was closed. It's clearly a bug where the first activation is not handled correctly when AD is started between sunset and the offset time.

acockburn:
Our bot closed it automatically due to inactivity, it is indeed a bug and we will get to it

ekutner:
This issue was fixed and merged months ago. I'll close it as completed now.

PR #2412, opened 2025-08-13
Fix handling of run_at_sunset/sunrise with offset

ekutner:
This PR fixes issue #2188


acockburn:
Hi there - quick update in this - we do thank you for the fixes, but we have done some other work in this area that may conflict, or may already fix the issue, we are double checking.

ekutner:
Looking at the dev branch code I don't believe it's fixed. The problem is simple, the current code starts by always taking the next sunset/sunrise and building on top of that, however, when an offset is specified and the code is running after today's sunset but before the sunset+offset time, then looking at the next sunset is wrong. In that case you need to take today's sunset and then add the offset to it. So the fix is still required and merging it would be greatly appreciated because it's a problem I encounter all the time, having a lot of sunset+offset schedules.

acockburn:
After a review we agree - I'll merge this to the next release.

ekutner:
Thank you

jsl12:
@ekutner my apologies on this one - I was the holdup. This has been on my list to look into for a while, but other things kept coming up and I thought there was a chance this wasn't an issue any more.

We appreciate your patience

ekutner:
And I appreciate you're great work on this project. It allowed me to do things with Home Assistant that are unthinkable with the built in scripting engine. Thank you all again!

PR #2445, opened 2025-10-14
reload modified apps on SIGUSR2

chatziko:
This tiny PR makes AppDaemon check for modified apps (and reload them) when SIGUSR2 is received.

This is similar to SIGHUP (which reloads all apps,) but only reloads modified ones. It is useful in `production_mode` in which modification checks do not happen automatically. So one can upload the code, send SIGUSR2 and have AppDaemon quickly reload only what's needed.


acockburn:
Thanks!

PR #2465, opened 2025-11-02
Add request context logging for failed HASS API calls

ekutner:
Currently when a call to Home Assistant fails the log entry in appdaemon is pretty useless. It just says "Error with websocket result" and gives the error information but without knowing the context of the call, the service that was called, it's impossible to know what user code caused it in the first place. Theoretically it may be possible to get this information by running appdaemon with loglevel=DEBUG but that is not practical for a production environment.

This PR saves the request based on the call_id, in the same way that silent_result is handled, and then writes it to the log in case the response indicated an error.

### DIFF
--- a/appdaemon/plugins/hass/hassplugin.py
+++ b/appdaemon/plugins/hass/hassplugin.py
@@ -80,6 +80,7 @@ class HassPlugin(PluginBase):
 
     _result_futures: dict[int, asyncio.Future]
     _silent_results: dict[int, bool]
+    _request_context: dict[int, dict[str, Any]]
     startup_conditions: list[StartupWaitCondition]
     maintenance_tasks: list[asyncio.Task]
     """List of tasks that run in the background as part of the plugin operation. These are tracked because they might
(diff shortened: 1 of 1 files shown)


jsl12:
Thanks for the contribution, and apologies for the delay. I've been pretty busy recently, and I assume @acockburn has been too.

I really like the idea of including contextual information with each request, but I think maybe we should make each context a dataclass and fold the other stuff (`result_futures` and `_silent_results`) into it too.

In the meantime, there are some [other ways](https://appdaemon.readthedocs.io/en/latest/HASS_API_REFERENCE.html#debugging) to debug wonky service calls. Also, the standard error messages have been improved relatively recently, so it should pass through whatever the error message is directly from HA

ekutner:
It would definitely be nicer but I was going for a minimal change and followed a pattern that was already in place for connecting requests with responses.
The "other ways" you mention are useful when you know where the problem is and just need more information about it, but they're not that useful when there are many service calls and you get some generic statement from HA that doesn't actually help with understanding which call is causing it. In such a case it is very useful to know exactly which service call ended up with the error response. In my case, adding this context information was the only way I could find the source of an error I was getting because the debug log was just too messy to follow (I have a rather complex implementation).

PR #2473, opened 2025-11-07
Fix reading of http password and ha_key

### LINKED ISSUE #2462: Can't connect to api when password is set
### What happened?

As described [here](https://appdaemon.readthedocs.io/en/latest/APPGUIDE.html#api-security) I have configured an app to react to being called via the `api`. I am using it as a webhook for uptime kuma so that I can get an alert when appdaemon stops responding.

I have this in my `configuration.yaml`
```
appdaemon:
  plugins:
    HASS:
      type: hass
      ha_url: "http://homeassistant.lan:8123"
      token: !secret token
  time_zone: Europe/Berlin
  elevation: 1
  production_mode: False
  namespaces:
    custom_services:
      #writeback: performance
      writeback: hybrid
http:
  url: http://0.0.0.0:5050 
  password: fred
api:
hadashboard:

```
When the password is commented out it works as expected:

```
homeassistant:~/app# curl -H "x-ad-access: fred" http://homeassistant.lan:5050/api/appdaemon/uptime_kuma_endpoint
{"message": "works"}homeassistant:~/app#
```

But it does not work when a password is set:

```
homeassistant:~/app# curl -H "x-ad-access: fred" http://homeassistant.lan:5050/api/appdaemon/uptime_kuma_endpoint
<html><head><title>401 Unauthorized</title></head><body><h1>401 Unauthorized</h1>Error in API Call</body></html>homeassistant:~/app#
```

### Version

4.5.12

### Installation type

Docker container

### Relevant log output

```sh

```

```sh

```

### Anything else?

I noticed that the doc says this

`A security key is added for the API with the api_key directive described in the [Installation Documentation](https://appdaemon.readthedocs.io/en/latest/INSTALL.html)`

But the linked documentation doesn't contain the key `api_key` so I am not sure whether I am doing something wrong or whether the documentation is missing some info.

cebtenzzre:
AppDaemon 4.5.0 migrated the configuration to use Pydantic models. These models use SecretStr for fields that contain secrets.

This seems to have caused two regressions, because of the automatic obfuscation that converts `str(SecretStr('abc'))` -> `'**********'`.

1. Password login became impossible, because HTTP.password came via model_dump(), which obfuscates secrets.
2. `x-ha-access` was no longer set correctly, because it did not unwrap `ha_key` before using it.

This is fixed by calling get_secret_value() on the SecretStr in both cases.

I have confirmed that password login works again after this fix.

Fixes #2462

### DIFF
--- a/appdaemon/http.py
+++ b/appdaemon/http.py
@@ -162,6 +162,8 @@ class HTTP:
         self.static_dirs = {}
 
         self._process_http(self.http)
+        if (cfg := main_cfg.http) is not None and (p := cfg.password) is not None:
+            self.password = p.get_secret_value()  # read secret directly
 
         self.app_endpoints = {}
         self.app_routes = {}
@@ -386,7 +388,6 @@ class HTTP:
         self.setup_dashboard_routes()
 
     def _process_http(self, http):
-        self._process_arg("password", http)
         self._process_arg("tokens", http)
         self._process_arg("work_factor", http)
         self._process_arg("ssl_certificate", http)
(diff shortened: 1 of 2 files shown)


acockburn:
Thanks!

PR #2475, opened 2025-11-08
only clean query parameters

### LINKED ISSUE #2464: Regression on set_state api call
### What happened?

Hello

I am updating some small home assistant script  that were previously running smoothly in 4.4.x 

When I use the set_state api call, the calls are successful but  it messes up with the boolean that are passed as attributes:

- attributes set to False are dropped and will not appear in the entity 
- attributes set to True are changed to a string "true" in home assistant

This was not the case in the 4.4.x where all attributes would be transmitted with the correct type

Look like this might be linked to the work done in #2425 

### Version

4.5.12

### Installation type

Home Assistant add-on

### Relevant log output

```sh
Nothing
```

```sh
attributes = {
        "friendly_name": "Octopus Agile Rates",
        "unit_of_measurement": "GBP/kWh",
        "plunge": False,
        "peak": True,
    }
    self.set_state(
        "event.test_entity",
        state=1,
        attributes=attributes,
        replace = True,
    )
```

### Anything else?

Doing a quick hatchet job in utils.py (as per below), I can obtain the correct behavior, but this would likely fixed the other bug you were addressing

```
def clean_kwargs(val: Any, *, http: bool = False) -> Any:
    match val:
        #case True if http:                                                                                                                                                  
        #    return "true"                                                                                                                                                   
        case str() | int() | float() | bool() | None:
            return val
        case datetime():
            return val.isoformat()
        case Mapping():
            return {k: clean_kwargs(v, http=http) for k, v in val.items()}
        case Iterable():
            return [clean_kwargs(v, http=http) for v in val]
        case _:
            return str(val)

def clean_http_kwargs(val: Any) -> Any:
    """Recursively cleans the kwarg dict to prepare it for use in HTTP requests."""
    cleaned = clean_kwargs(val, http=True)
    pruned = remove_literals(cleaned, (None,)) # Dropped False from the list
    return pruned
```

cebtenzzre:
There is no need to "clean" the json body of a POST request to alter things like booleans - JSON already has a way to represent them. That wasn't done in 4.4.x, which had a more transparent set_state.

One possible exception is datetime, which doesn't have one true JSON representation. But we can enhance convert_json to always use isoformat.

(Also, DELETE shouldn't have a body - but it may have query params. We don't currently use kwargs with DELETE so this isn't crucial, but it's important for correctness.)

Fixes #2464

### DIFF
--- a/appdaemon/utils.py
+++ b/appdaemon/utils.py
@@ -24,13 +24,14 @@ from logging import Logger
 from pathlib import Path
 from time import perf_counter
 from typing import TYPE_CHECKING, Any, Callable, Coroutine, Literal, ParamSpec, Protocol, TypeVar
+from urllib import parse
 
 import dateutil.parser
 import tomli
 import tomli_w
 import yaml
 from astral.location import Location
-from pydantic import BaseModel, ValidationError
+from pydantic import AnyHttpUrl, BaseModel, ValidationError
 from pytz import BaseTzInfo
 
 from appdaemon.parse import parse_datetime
@@ -1172,13 +1173,11 @@ def clean_http_kwargs(val: Any) -> Any:
     return pruned
 
 
-def make_endpoint(base: str, endpoint: str) -> str:
+def make_endpoint(base: str | AnyHttpUrl, endpoint: str) -> str:
     """Formats a URL appropriately with slashes"""
-    if not endpoint.startswith(base):
-        result = f"{base}/{endpoint.strip('/')}"
-    else:
-        result = endpoint
-    return result.strip("/")
+    parsed = parse.urlparse(str(base))
+    result = parsed._replace(path=endpoint)
+    return parse.urlunparse(result)
 
 
 def unwrapped(func: Callable) -> Callable:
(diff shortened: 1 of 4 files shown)


jsl12 (approved):
Using urllib to actually parse things should resolve this stuff about where the slashes are, so I don't think #2485 is necessary any more

cebtenzzre:
Well, that PR is about websocket_url and states_api in [plugin.py](https://github.com/cebtenzzre/appdaemon/blob/dev/appdaemon/models/config/plugin.py), and this PR doesn't even touch that file. Are you looking for a urllib-based version of #2485?

appdaemon/utils.py
```diff
@@ -1167,13 +1173,11 @@ def clean_http_kwargs(val: Any) -> Any:
...
+def make_endpoint(base: str | AnyHttpUrl, endpoint: str) -> str:
     """Formats a URL appropriately with slashes"""
-    if not endpoint.startswith(base):
-        result = f"{base}/{endpoint.strip('/')}"
-    else:
-        result = endpoint
-    return result.strip("/")
+    parsed = parse.urlparse(str(base))
+    result = parsed._replace(path=endpoint)
+    return parse.urlunparse(result)
```
cebtenzzre: I think this breaks base URLs with a path (e.g. `http://supervisor/core`) - it will replace `/core` with the endpoint path instead of appending. See #2485 for a better fix using yarl.

PR #2482, opened 2025-11-11
Enhance Docker tutorial with compose flags note

mmattel:
Added note about using flags with docker compose in the troubleshooting section.


acockburn:
Thanks!

PR #2489, opened 2025-11-23
remove match statement in get_history

cebtenzzre:
I'm not sure that ClientResponse can actually be matched like this.

It shouldn't be necessary here anyway - returning None implicitly if resp is not a ClientResponse doesn't seem like the right thing to do here.

### DIFF
--- a/appdaemon/plugins/hass/hassplugin.py
+++ b/appdaemon/plugins/hass/hassplugin.py
@@ -14,7 +14,7 @@ from time import perf_counter
 from typing import Any, Literal, Optional
 
 import aiohttp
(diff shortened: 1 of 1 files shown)


cebtenzzre:
cc @jsl12

jsl12:
This does "work" in the sense that you _can_ use match/case to unpack objects like that, but I agree with you here.

I think I had that in there to force pyright to recognize the type or something, but it seems unnecessary now and your way is cleaner.

jsl12 (approved):

cebtenzzre:
You're right that this works, I just tested it. Still, the point of this match statement looks like to fall through if the ClientResponse doesn't match, which is confusing.

PR #2502, opened 2025-12-14
Generic ADAPI Class, for specialized AppConfig

TCampmany:
This PR makes the ADAPI generic so apps can declare a specialized, typed configuration model, and access validated settings while keeping the legacy dict for compatibility.

On startup, configuration is re-validated/converted into the declared model type, and mismatches are logged and prevent the app from starting (fail fast).

The same typed configuration support is propagated to the Home Assistant and MQTT APIs, with docs updated to show how to use typed app configs. `config_model``args`

### DIFF
--- a/appdaemon/adapi.py
+++ b/appdaemon/adapi.py
@@ -12,7 +12,7 @@ from copy import deepcopy
 from datetime import timedelta
 from logging import Logger
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Literal, TypeVar, overload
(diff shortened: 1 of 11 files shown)


jsl12 (approved):
Very cool! I think this will be really useful

hugh-martin:
Nice!

jsl12 (changes requested):
Just needs to pass the tests

jsl12 (approved):

acockburn:
Hu there - thanks for this! We have merge conflicts, could you take a look and fix, then we can merge - thanks.

PR #2515, opened 2025-12-28
parse: fix time zone handling in scheduler time parsing

### LINKED ISSUE #2338: Handling of datetime.tim() when scheduling run_daily() changed between 4.4.2 and 4.5.5
### What happened?

I have a handful of daily tasks scheduled using `run_daily()`. Those tasks were set like this:

```
self.ad.run_daily(self.turn_lights, time(23, 59, 0), state="OFF")
```

Where `time(23, 59, 0)` used to result in the callback being scheduled for 11:59 p.m. local time. Since changing to 4.5.5, that is running at 11:59 p.m. UTC instead.

Switching the call to use a string yields the desired behavior:

```
self.ad.run_daily(self.turn_lights, "23:59:00", state="OFF")
```

I've changed all my calls to use the string parameter, so it no longer affects me. I can see an argument for either behavior, but if left as it currently is it would constitute a breaking change.

### Version

4.5.5

### Installation type

Docker container

### Relevant log output

```sh

```

```sh

```

### Anything else?

_No response_

### LINKED ISSUE #2388: Timezone is handled wrong
### What happened?

I was running into an issue when I pass a time object without timezone information into adapi.run_daily. my timezone is Europe/Vienna - so I'm currently 2h ahead GMT. However, scheduler would schedule the callbacks 55min lates - so for example instead of 23:00 it would run on 23:55.
I think I found the bug in the recently added timezone handling in [adapi.run_daily](https://github.com/AppDaemon/appdaemon/blob/dev/appdaemon/adapi.py#L3061C28-L3065C57):
``` 
                if start.tzinfo is None:
                    start = start.replace(tzinfo=self.AD.tz)
                date = await self.date()
                start = dt.datetime.combine(date, start)

```
With that code the time is calulated based on LMT not GMT

I wrote simple test script to highlight the issue:
```
>>> from datetime import datetime
>>> import pytz
>>> tz = pytz.timezone("Europe/Vienna")
>>> time = datetime.strptime("23:00", '%H:%M').time()
>>> time = time.replace(tzinfo = tz)
>>> date = datetime.now().date()
>>> datetime.combine(date, time)
datetime.datetime(2025, 6, 27, 23, 0, tzinfo=<DstTzInfo 'Europe/Vienna' LMT+1:05:00 STD>)
```

I refactored the code a bit to use localize from pytz:

``` 
>>> from datetime import datetime
>>> import pytz
>>> tz = pytz.timezone("Europe/Vienna")
>>> time = datetime.strptime("23:00", '%H:%M').time()
>>> date = datetime.now().date()
>>> start = datetime.combine(date, time)
>>> tz.localize(start)
datetime.datetime(2025, 6, 27, 23, 0, tzinfo=<DstTzInfo 'Europe/Vienna' CEST+2:00:00 DST>)
``` 
so it seems that combining date and time before doing tz.localize gives the correct result, while time.replace(tzinfo=tz) before combining date and time results in a wrong time. 

Therefore I think that the code in adapi.run_daily needs to be refactored to something like this:
```
                has_no_tzinfo = start.tzinfo is None
                date = await self.date()
                start = dt.datetime.combine(date, start) 
                if has_no_tzinfo:
                    start = self.AD.tz.localize(start)
```

Let me know what you think,
Thanks, Roman

### Version

4.5.11

### Installation type

Docker container

### Relevant log output

```sh

```

```sh

```

### Anything else?

_No response_

### LINKED ISSUE #2391: Schedy fails to operate properly on v4.5.11, works fine on v4.5.8
### What happened?

The HASS APP Shchedy runs properly and sets the thermostat correctly at startup, however, any time based changes in the Schedy app fail to apply. 

### Version

4.5.11

### Installation type

Home Assistant add-on

### Relevant log output

```sh
I have info logging enabled, and nothing shows up at the time of the scheduled change.
```

```sh
thermostat_schedule:
  module: hass_apps_loader
  class: SchedyApp
  debug: true
  actor_type: thermostat
  schedule_snippets:

######################
### Cooling Season ###
######################

    cooling_season_home:
    - v: 75
      rules:                           
      - { start: "7:00", end: "21:00"}
    - v: 69

    cooling_season_vacation:
    - v: 78

    cooling_season_guest:                               
    - v: 75                                             

######################
### Heating Season ###
######################

    heating_season_home:
    - v: 70
      rules:
      - {start: "07:00", end: "21:00"}
    - v: 67

    heating_season_vacation:
    - v: 67

    heating_season_guest:
    - v: 70
      rules:                                            
      - {start: "07:00", end: "21:00"} 
    - v: 67

  rooms:
    whole_house:
      actors:
        climate.thermostat_main:
          supports_hvac_modes: false
      schedule:
        ######################
        ### Cooling Season ###
        ######################        
        - x: "IncludeSchedule(schedule_snippets['cooling_season_home']) if state('climate.thermostat_main') == 'cool' and state('input_select.thermostat_schedule_main') == 'Home' else Next()"
        - x: "IncludeSchedule(schedule_snippets['cooling_season_vacation']) if state('climate.thermostat_main') == 'cool' and state('input_select.thermostat_schedule_main') == 'Vacation' else Next()"
        ######################
        ### Heating Season ###
        ######################
        #- x: "IncludeSchedule(schedule_snippets['heating_season_guest']) if state('climate.thermostat_main') == 'heat' and is_on('switch.guest_mode') else Next()"
        - x: "IncludeSchedule(schedule_snippets['heating_season_home']) if state('climate.thermostat_main') == 'heat' and state('input_select.thermostat_schedule_main') == 'Home' else Next()"
        - x: "IncludeSchedule(schedule_snippets['heating_season_vacation']) if state('climate.thermostat_main') == 'heat' and state('input_select.thermostat_schedule_main') == 'Vacation' else Next()"

  watched_entities:
    - climate.thermostat_main
    - input_select.thermostat_schedule_main
    - switch.guest_mode
```

### Anything else?

_No response_

### LINKED ISSUE #2393: run_daily does not work..
### What happened?

does now work in 4.x and works in 3.0.5
```
        self.run_daily(self._cb_event, 
                       self.parse_time("16:50:00"), 
                       constrain_days=ada.schedule.day_of_week(5),
                       a="1",b="2")
```

does now work in 4.x and works in 3.0.5

```
     self.run_daily(self._cb_event, 
                       self.parse_time("16:50:00"), 
                       a="1",b="2")
```

back to 3.0.5 

### Version

4.5.11

### Installation type

Docker container

### Relevant log output

```sh

```

```sh

```

### Anything else?

_No response_

### LINKED ISSUE #2400: Unexpected 36-Minute Offset in run_daily Scheduler on Raspberry Pi 5 (HAOS, Python 3.12, ARM64)
### What happened?

**Describe the bug**
There is a consistent and repeatable 36-minute offset when scheduling callbacks using run_daily() in AppDaemon on Raspberry Pi 5 with Home Assistant OS (HAOS).
The system time, Home Assistant time, and AppDaemon logs all show the correct time, but the callback scheduled by run_daily always fires exactly 36 minutes later than the time specified in the YAML configuration.

**Environment**
- Hardware: Raspberry Pi 5 (64-bit, ARM)
- Home Assistant OS version: 16.0
- AppDaemon version: 4.5.11 (official addon)
- Python version (as in AppDaemon logs): 3.12.11
- Home Assistant Core version: 2025.7.2
- System time: Confirmed correct with date and in AppDaemon logs

**How to reproduce the issue**
1. Install AppDaemon 4.x as an official add-on on Home Assistant OS running on a Raspberry Pi 5.
2. Create a simple app, e.g.:

          import datetime
      
          class LightTimer(hass.Hass):

              def initialize(self):
                  self.on_time = self.args.get("on_time", "10:00:00")
                  parsed_on = self.parse_time(self.on_time)
                  self.log(f"Parsed on_time: {parsed_on}")
                  self.run_daily(self.turn_on_light, parsed_on)
      
              def turn_on_light(self, kwargs):
                  self.log("Light should turn on now.")

3. Configure apps.yaml:

               light_timer:
                  module: light_timer
                  class: LightTimer
                  light: light.xyz
                  on_time: "10:00:00"

4. Restart AppDaemon.
5. Observe that in the AppDaemon UI ("Scheduler Callbacks"), the scheduled time for the callback is always 10:36:00 when "10:00:00" is set, and so on (always +36 minutes).

**Observed behavior**
- The scheduled callback is always +36 minutes later than the requested time, regardless of the value specified.
- The offset is always exactly 36 minutes (e.g. 09:10:00 becomes 09:46:00, 12:12:00 becomes 12:48:00).
- All system, Home Assistant, and AppDaemon times are correct and in sync.
- No stubs, duplicates, or timezone inconsistencies in configuration.

**Expected behavior**
- run_daily should schedule the callback at the time specified (e.g. 10:00:00).

**Workaround**
Currently, I am manually subtracting 36 minutes in my app code:

              import datetime
              parsed_on = self.parse_time(self.on_time)
              fixed_on = (datetime.datetime.combine(datetime.date.today(), parsed_on) - datetime.timedelta(minutes=36)).time()
              self.run_daily(self.turn_on_light, fixed_on)

But this is just a workaround for a deeper issue.

**Additional info**
- Full restart of the system does not resolve the issue.
- The problem is present on Raspberry Pi 5, Home Assistant OS 16.0, Python 3.12.11.
- The same code on older hardware or with older versions of AppDaemon/Python does not show this bug.
- No error messages, just the consistent scheduling offset.

### Version

4.5.11

### Installation type

Home Assistant add-on

### Relevant log output

```sh

```

```sh

```

### Anything else?

_No response_

cebtenzzre:
### Problem

Users reported that `run_daily()` and other scheduler functions were scheduling callbacks at incorrect times, sometimes with unexpected offsets from the expected time. This regression was introduced after v4.4.2.

**Example from issue:**
```python
run = datetime.time(8, 0, 0)
self.run_daily(self.run_daily_c, run, val='on')
```
Expected execution: `08:00:00`  
Actual execution: `09:07:00`

### What Caused the Regression

In v4.4.2, the code correctly handled pytz timezones:

```python
# Old approach (correct)
event = datetime.combine(today, when)          # 1. Create NAIVE datetime
aware_event = self.AD.sched.convert_naive(event)  # 2. Localize with tz.localize()
```

The `convert_naive()` method properly used `self.AD.tz.localize(dt)`, which is the required pattern for pytz timezones to correctly determine the UTC offset based on DST for that specific date.

When the time parsing was refactored into `parse.py`, this pattern was replaced with:

```python
# New approach (broken)
result = datetime.combine(date, parsed_time, tzinfo=now.tzinfo)
```

This is **incorrect for pytz timezones**. Unlike standard `tzinfo` implementations, pytz timezones are stateful and require the `localize()` method to properly determine the UTC offset. When you pass a pytz timezone directly to `datetime.combine()` or use `dt.replace(tzinfo=...)`, pytz may apply an incorrect default offset that doesn't account for DST on that specific date.

### Solution

Instead of reverting to the old `localize()` pattern (which is pytz-specific and non-standard), we normalize pytz timezones to `ZoneInfo` (PEP 615) at the boundary:

1. **`normalize_tz(tz)`** - Converts pytz timezones to their equivalent `ZoneInfo` using the IANA zone name. ZoneInfo behaves like a standard `tzinfo` and handles DST correctly with normal `replace(tzinfo=...)` calls.

2. **`localize_naive(naive_dt, tz)`** - A helper that normalizes the timezone first, then uses standard `replace(tzinfo=...)`.

This approach:
- Centralizes the pytz→ZoneInfo conversion in one place
- Uses standard Python datetime semantics after normalization
- Properly handles DST by letting ZoneInfo interpret wall-clock times correctly
- Avoids pytz-specific API calls scattered throughout the codebase

Fixes #2338
Fixes #2388
Fixes #2391
Fixes #2393
Fixes #2400

### DIFF
--- a/appdaemon/parse.py
+++ b/appdaemon/parse.py
@@ -1,12 +1,55 @@
 import re
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
-from datetime import datetime, time, timedelta
+from datetime import datetime, time, timedelta, tzinfo
 from functools import partial
 from typing import ClassVar, Literal
+from zoneinfo import ZoneInfo
 
 import astral
+import pytz
 from astral.location import Location
+
+
+def normalize_tz(tz: tzinfo) -> tzinfo:
+    """Convert pytz timezone to ZoneInfo for clean stdlib-compatible handling.
+
+    pytz timezones don't behave like normal tzinfo implementations and require
+    special localize()/normalize() calls. By converting to ZoneInfo at the boundary,
+    we can use standard replace(tzinfo=...) and astimezone() everywhere else.
+
+    Args:
+        tz: Any tzinfo object (pytz, ZoneInfo, or fixed-offset)
+
+    Returns:
+        ZoneInfo if input was pytz with an IANA zone name, otherwise unchanged
+    """
+    if isinstance(tz, pytz.tzinfo.BaseTzInfo) and tz.zone is not None:
+        return ZoneInfo(tz.zone)
+    return tz
+
+
+def localize_naive(naive_dt: datetime, tz: tzinfo) -> datetime:
+    """Interpret a naive datetime as wall-clock time in the given timezone.
+
+    This normalizes pytz timezones to ZoneInfo first, so we can use standard
+    replace(tzinfo=...) semantics instead of pytz's localize().
+
+    Args:
+        naive_dt: A naive datetime (no tzinfo)
+        tz: The timezone to interpret the datetime in
+
+    Returns:
+        A timezone-aware datetime
+
+    Raises:
+        ValueError: If naive_dt already has tzinfo
+    """
+    if naive_dt.tzinfo is not None:
+        raise ValueError("expected naive datetime")
+    tz = normalize_tz(tz)
+    return naive_dt.replace(tzinfo=tz)
+
 
 CONVERTERS = {
     "hour": lambda v: timedelta(hours=v),
@@ -267,14 +310,16 @@ def resolve_time_str(
             else:
                 raise ValueError(f"Invalid time string: {input_string}")
 
+    tz = normalize_tz(now.tzinfo)
+
     offset = timedelta()
     match _parse(time_str):
         case time() as parsed_time:
-            result = datetime.combine(
+            naive_dt = datetime.combine(
                 (now + timedelta(days=days_offset)).date(),
                 parsed_time,
-                tzinfo=now.tzinfo
             )
+            result = localize_naive(naive_dt, tz)
         case datetime() as result:
             pass
         case Now(offset=offset):
@@ -338,14 +383,16 @@ def parse_datetime(
 
     assert isinstance(now, datetime) and now.tzinfo is not None, "Now must be a timezone-aware datetime"
 
+    tz = normalize_tz(now.tzinfo)
+
     offset = timedelta()
     match input_:
         case time() as input_time:
-            result = datetime.combine(
+            naive_dt = datetime.combine(
                 (now + timedelta(days=days_offset)).date(),
                 input_time,
-                tzinfo=now.tzinfo
             )
+            result = localize_naive(naive_dt, tz)
         case datetime() as result:
             result += timedelta(days=days_offset)
         case str() as time_str:
@@ -364,10 +411,9 @@ def parse_datetime(
 
     # Make the timezones match for the comparison below
     if result.tzinfo is None:
-        # Just adds the timezone without changing the time values
-        result = result.replace(tzinfo=now.tzinfo)
+        result = localize_naive(result, tz)
     else:
-        result = result.astimezone(now.tzinfo)
+        result = result.astimezone(tz)
 
     # The the days offset is negative, the result can't be forced to today, so set today to False
     if days_offset < 0:


cebtenzzre:
I think this fix can be a stepping stone toward a medium/long-term cleanup: **prefer `zoneinfo` internally**, while continuing to **accept pytz at the boundary** for backwards            compatibility.

#### Why move toward `zoneinfo`?

* **`pytz` does not implement “normal” `tzinfo` semantics**: it requires `localize()`/`normalize()`, and attaching a pytz zone via `replace(tzinfo=...)` is explicitly error-prone (this   regression is a concrete example of how easy it is to get wrong).
* **`zoneinfo` (PEP 615) is the stdlib timezone implementation**. It behaves like a regular `tzinfo`, so the “obvious” datetime operations do the right thing (`replace(tzinfo=...)`,      `astimezone()`, arithmetic).
* Centralizing on `zoneinfo` reduces the surface area for DST bugs and makes the code easier to reason about.
* It also reduces dependency complexity long-term (stdlib + optional `tzdata` on platforms that need it).

cebtenzzre:
The workaround SteLu1803 found in https://github.com/AppDaemon/appdaemon/issues/2391#issuecomment-3259404256 confirms the root cause. They discovered that passing a **string** to `run_daily` instead of a `datetime.time` object works correctly.

This makes sense because:

1. **Broken path** (using `datetime.time`):
   ```python
   self.run_daily(callback, datetime.time(8, 0, 0))  # ❌ Triggers buggy code
   ```
   This hit the `case time():` branch in `parse.py` that was incorrectly using `datetime.combine(..., tzinfo=pytz_tz)`.

2. **Working workaround** (using string):
   ```python
   parsed_time = self.app.parse_time("08:00:00", aware=True)
   self.app.run_daily(callback, parsed_time)  # ✅ Bypasses the bug
   ```
   When you pass a string, it goes through a different code path. And by passing `aware=True`, they get back a properly localized datetime that was already correctly converted.

The fix addresses this directly - after the fix, users can use `datetime.time` objects as intended without needing the string workaround. The fix ensures that when a `time` object comes in, it's properly converted by:
1. Creating a naive datetime first
2. Normalizing the pytz timezone to ZoneInfo  
3. Using standard `replace(tzinfo=...)` which works correctly with ZoneInfo

Issue #2417, opened 2025-09-04
Scheduler does not include offsets after callback time recalculation


kbuc:
### What happened?

I run AD as addon on HA. I schedule several callbacks around the sunset with the code:

```
for n, time_offset in self.positions.items():
	self.run_at_sunset(self.function, offset=time_offset, n=n)
```

Offsets are different negative values, about 40 seconds difference (like -100, -160, -200, -240,... etc). 

Afer AD first start or restart, the callbacks are properly scheduled, but after the sunset when things are recalculated for the next day, all the callbacks register at the same time (sunset with no offset, I guess). See the attached screenshots. 

Left: automatic recalculation after firing the previous day. Everything is wrongly scheduled at the same moment ignoring offsets. Right: after restart, the same calculation is done properly with offsets.

<img width="252" height="545" alt="Image" src="https://github.com/user-attachments/assets/bb40fb10-4bb4-4f97-a625-b9c72d11e6a6" />
<img width="213" height="508" alt="Image" src="https://github.com/user-attachments/assets/5d7a49e9-1644-4e86-b357-2d04e83ebf23" />

### Version

4.5.11

### Installation type

Home Assistant add-on

### Relevant log output

```sh

```

### Relevant code in the app or config file that caused the issue

```sh

```

dekiesel:
I am wondering if this is related to my issue in #2491.

PR #2518, opened 2025-12-28
scheduler: improve handling of offsets

### LINKED ISSUE #2384: run_every with start="now" doesn't run immediately
### What happened?

With the introduction of version 4.5.0 the [run_every](https://appdaemon.readthedocs.io/en/latest/AD_API_REFERENCE.html#appdaemon.adapi.ADAPI.run_every) function changed in behavior regarding the `start` parameter

before 4.5.0:

> start – A Python datetime object that specifies when the initial callback will occur, or can take the now string alongside an added offset. If given in the past, it will be executed in the next interval time.

when passing `"now"` for `start` and e.g. a 10 minute `interval`, the callback would run immediately, and then 10 minutes later, another 10 minutes later, etc.

4.5.0+:

> > start (str, datetime.time, datetime.datetime, optional) –
> Start time for the interval calculation. If this is in the future, this will be the first time the callback is triggered. If this is in the past, the intervals will be calculated forward from the start time, and the first trigger will be the first interval in the future.
> 
>     If this is a str it will be parsed with parse_time().
> 
>     If this is a datetime.time object, the current date will be assumed.
> 
>     If this is a datetime.datetime object, it will be used as is.

If you pass the same `"now"` for `start` and a 10 minute `interval`, the first callback run would be in 10 minutes, and then 10 minutes after that, etc. Basically missing the expected immediate run.

### Version

4.5.0+

### Installation type

Home Assistant add-on

### Relevant log output

```sh

```

```sh

```

### Anything else?

_No response_

cebtenzzre:
These changes are inspired by #2417 to generally improve the code quality in this area.

This attempts to fix two potential issues (which I haven't made an effort to reproduce):
- passing an offset greater than the interval might cause the scheduler to do weird things later - now it throws an error immediately
- if random_start is use with a sun event, but the base offset is zero, it might schedule the event to occur again the same day - now it explicitly schedules based on the *next* sun event in either case

Fixes #2384

### DIFF
--- a/appdaemon/adapi.py
+++ b/appdaemon/adapi.py
@@ -2748,7 +2748,7 @@ class ADAPI:
         return await self.AD.sched.reset_timer(self.name, handle)
 
     @utils.sync_decorator
-    async def info_timer(self, handle: str) -> tuple[dt.datetime, int, dict] | None:
+    async def info_timer(self, handle: str) -> tuple[dt.datetime, float, dict] | None:
         """Get information about a previously created timer.
 
         Args:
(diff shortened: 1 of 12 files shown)


cebtenzzre:
I just added commit [6bef390](https://github.com/AppDaemon/appdaemon/pull/2518/changes/6bef390119f08c916f1ea004f3fa8021ecb43a1d) which should fix #2384.

PR #2526, opened 2026-01-08
patch for get_next_period and start  == "now"

jsl12:
This should fix chatziko's semantic issue


chatziko:
I'd vote for `start="immediate"`, instead of adding new arguments (with `"now"`, `None` keeping the old semantics).

PR #2594, opened 2026-04-11
Fix set_state dropping False, 0, and None values

### LINKED ISSUE #2464: Regression on set_state api call

### LINKED ISSUE #2492: clean_http_kwargs strips zero values (0, 0.0) from nested dicts due to 0 == False in Python
### What happened?

The clean_http_kwargs function in appdaemon/utils.py incorrectly removes integer 0 and float 0.0 values from nested dictionaries when preparing HTTP requests. This is because it uses remove_literals(cleaned, (None, False)) and in Python, 0 == False and 0.0 == False both evaluate to True.

**Steps to Reproduce:**
`from appdaemon.utils import clean_http_kwargs
import json

test_data = {
    'state': '0.08',
    'attributes': {
        'prices': {
            '2025-11-29T00:00:00+02:00': {'price': 0.0, 'intervals': 4},
            '2025-11-29T11:00:00+02:00': {'price': 0.08, 'intervals': 4}
        }
    }
}

result = clean_http_kwargs(test_data)
print(json.dumps(result, indent=2))`

**Expected Result:**
`{
  "state": "0.08",
  "attributes": {
    "prices": {
      "2025-11-29T00:00:00+02:00": {"price": 0.0, "intervals": 4},
      "2025-11-29T11:00:00+02:00": {"price": 0.08, "intervals": 4}
    }
  }
}`

**Actual Result:**
`{
  "state": "0.08",
  "attributes": {
    "prices": {
      "2025-11-29T00:00:00+02:00": {"intervals": 4},
      "2025-11-29T11:00:00+02:00": {"price": 0.08, "intervals": 4}
    }
  }
}`

The price key is completely missing from the first entry.

### Version

4.5.12

### Installation type

Docker container

### Relevant log output

```sh

```

```sh
In appdaemon/utils.py, the remove_literals function (line ~1102) uses equality comparison:

`def remove_literals(val: Any, literal: Sequence[Any]) -> Any:
    """Remove instances of literals from a nested data structure."""
    match val:
        case str():
            return val
        case Mapping():
            return {k: remove_literals(v, literal) for k, v in val.items() if v not in literal}
        # ...`

The clean_http_kwargs function calls it with (None, False):
`
def clean_http_kwargs(val: Any) -> Any:
    cleaned = clean_kwargs(val, http=True)
    pruned = remove_literals(cleaned, (None, False))
    return pruned
`
The problem is that v not in literal uses == comparison, and in Python:

0 == False → True
0.0 == False → True
0 in (None, False) → True
0.0 in (None, False) → True
Suggested Fix:

Use identity comparison (is) instead of equality (==) in remove_literals
```

### Anything else?

_No response_

### LINKED ISSUE #2531: Regression on set_state api call
### What happened?

A little while back I posted about a regression with the set_state api call. (issue #2464 )

Whilst some changes were made the issue is still here for me. Every time I set a state, every attribute that is equal to False or 0 will be ignored

I created a small applet below 

If you run that code with version 4.5.11, you get an entity with the following attributes

```
rate: 0
friendly_name: Test Entity
unit_of_measurement: GBP/kWh
plunge: false
plunge_start: false
```

If you run it with 4.5.12 or 4.5.13 you get the following attributes

```
friendly_name: Test Entity
unit_of_measurement: GBP/kWh
```

### Version

4.5.13

### Installation type

Python virtual environment

### Relevant log output

```sh

```

```sh
Test.yaml

test_case:
  module: test
  class: TestApp



Test.py

class TestApp(hass.Hass):

    def initialize(self):
        self.log("run test app")
        attributes = {
            "rate": 0,
            "friendly_name": "Test Entity",
            "unit_of_measurement": "GBP/kWh",
            "plunge": False,
            "plunge_start": False,
        }

        self.set_state(
            "sensor.test_case_sensor",
            1,
            attributes=attributes,
            replace = True,
        )


Config file

ppdaemon:
  time_zone: Europe/London
  latitude: 51
  longitude: 0
  elevation: 20
  plugins:
    HASS:
      type: hass
      ha_url: http://localhost:8123
      token: somethingsomething
```

### Anything else?

_No response_

Ten0:
Fixes #2531
Fixes #2464
Fixes #2492

Two bugs caused set_state (and other POST calls) to lose attribute values:

1. remove_literals used equality (==) to filter values, so 0 and 0.0 were removed alongside False (because 0 == False in Python). Fixed by switching to identity comparison (is).

2. clean_http_kwargs was applied globally to all HTTP methods, stripping False, None, and converting True to "true" even in POST JSON bodies where these are valid values. Replaced with a new self-contained clean_http_params_for_urlencode applied only to GET/DELETE query parameters; POST bodies are now passed transparently through the session's convert_json serializer. Removed the now-unused clean_kwargs function.

Also:
- Removed redundant clean_kwargs + remove_literals from websocket path; convert_json handles datetime serialization.
- Performance tracking now computes actual payload size per HTTP method.

I have tested this on my own HomeAssistant instance, and this does indeed solve the issue and it seems my home finally works properly without triggering exceptions on 4.5 with that change, so I might finally update from 4.4 once this is released. 🎉


jsl12 (approved):

acockburn:
Hi there - thanks for this! We have merge conflicts, if you could take a look and fix we can get this merged - thanks!

Ten0:
I checked not a week ago and there were none... 😕

Ten0:
Done!

acockburn:
Thanks!

PR #2597, opened 2026-04-13
perf: cut per-event overhead in _sub_stack and events.process_event

jkempson:
Two small performance fixes for AppDaemon's per-event path.

## 1. `ADAPI._sub_stack`: skip `inspect.stack()` when no placeholders are present

Today `_sub_stack` calls `inspect.stack()` unconditionally on every string log, then checks whether `__module__` / `__line__` / `__function__` are actually in the message. Since most log calls don't use those placeholders, the stack walk is wasted work, and it's not cheap. `inspect.stack()` is O(frame count) plus source file reads via `linecache` per frame, and real AppDaemon callback stacks are 30 to 60 frames deep (asyncio, threadpool, dispatch, app).

This PR reorders the check: cheap substring test first, stack walk only when a placeholder is actually present. Behaviour is identical in both branches.

### Measurements

Measured inside a running AppDaemon 4.5.13 pod (Python 3.12, warm linecache):

| Stack depth | `inspect.stack()` cost |
|---:|---:|
| 8 | 8.9 ms |
| 18 | 10.6 ms |
| 33 | 13.8 ms |
| 63 | 33.9 ms |

Cold calls and GIL contention push it higher still.

### Real-world impact

In a motion-activated lighting deployment (around 10 zones, heavy `self.log` usage in callbacks), zone updates were taking 400 to 500 ms end to end. After bypassing `_sub_stack` for placeholder-free messages via a local monkey-patch equivalent to this change, the same updates run in 2 to 13 ms with nothing else modified.

### Change

- `adapi.py`: early return for non-strings, then for strings with no placeholders. Otherwise the existing stack walk and substitutions run unchanged at the same frame index (`stack[2]`).
- `tests/unit/test_sub_stack.py`: tests for non-string input, plain strings, each placeholder individually, and multiple placeholders in one message.

---

## 2. `Events.process_event`: skip the deepcopy when no `/stream` clients are connected

`Events.process_event` unconditionally deep-copies every inbound HA event when `http:` is enabled, so the copy can be forwarded to `/stream` for admin UI / HADashboard subscribers. The stream handler (`ADStream.process_event`) already short-circuits when `handlers` is empty, but the `deepcopy` at `events.py:272/275` happens *before* that check, so the cost is paid unconditionally.

For any install that enables `http:` for its REST endpoints (`/api/appdaemon/*`) but doesn't keep an admin UI / HADashboard tab open, the deepcopy is pure waste.

This PR adds an early return when `http.stream.handlers` is empty, before the `state_changed` equality check and before either `deepcopy()`. The existing path runs unchanged whenever at least one client is subscribed, so opening the admin UI restores full stream behaviour automatically.

### Measurements

py-spy (50 Hz sampling, 5-minute window) against a pod running ~150 apps with roughly 5 HA `state_changed` events/sec, no admin UI client connected:

| Sampled frame | Before | After |
|---:|---:|---:|
| `deepcopy` (copy.py) | 24.2% | <1% |
| `_deepcopy_dict` (copy.py) | 20.6% | <1% |
| `events.process_event` total | 12.0% | ~3% |

Combined impact on container CPU: sustained average dropped from ~443 mCPU to ~29 mCPU post-restart in our deployment (admin-loop fix also contributed — the deepcopy elimination is the larger share).

### Real-world impact

`/stream` subscribers are rare in normal operation — most deployments open the admin UI to debug and then close it. When nothing is subscribed, this branch was the dominant steady-state cost of AppDaemon in our environment.

### Change

- `events.py`: inside the `if self.AD.http is not None:` branch, check `http.stream.handlers`. Return early if empty. Otherwise the existing path (state-changed equality check, deepcopy, `stream_update`) runs unchanged.
- `tests/unit/test_event_stream_shortcircuit.py`: verifies `stream_update` is skipped when `handlers` is empty, called when it has clients, and that `http=None` still works.

I only touched that one block in `process_event` to keep the diff minimal.

### DIFF
--- a/appdaemon/adapi.py
+++ b/appdaemon/adapi.py
@@ -101,14 +101,21 @@ class ADAPI:
     @staticmethod
     def _sub_stack(msg):
         # If msg is a data structure of some type, don't sub
-        if isinstance(msg, str):
-            stack = inspect.stack()
-            if msg.find("__module__") != -1:
-                msg = msg.replace("__module__", stack[2][1])
-            if msg.find("__line__") != -1:
-                msg = msg.replace("__line__", str(stack[2][2]))
-            if msg.find("__function__") != -1:
-                msg = msg.replace("__function__", stack[2][3])
+        if not isinstance(msg, str):
+            return msg
+        # Fast path: avoid the expensive inspect.stack() call (which walks
+        # every frame and reads source lines via linecache) when no
+        # placeholders are present. The vast majority of log calls do not
+        # use __module__/__line__/__function__ substitution.
+        if "__module__" not in msg and "__line__" not in msg and "__function__" not in msg:
+            return msg
+        stack = inspect.stack()
+        if "__module__" in msg:
+            msg = msg.replace("__module__", stack[2][1])
+        if "__line__" in msg:
+            msg = msg.replace("__line__", str(stack[2][2]))
+        if "__function__" in msg:
+            msg = msg.replace("__function__", stack[2][3])
         return msg
 
     def _get_namespace(self, **kwargs):
--- a/appdaemon/events.py
+++ b/appdaemon/events.py
@@ -264,6 +264,15 @@ class Events:
             #
 
             if self.AD.http is not None:
+                # Short-circuit when nobody is subscribed to /stream. The
+                # deepcopy below exists only to feed stream_update, and
+                # ADStream.process_event is itself a no-op when handlers
+                # is empty. When no clients are connected this branch was
+                # the dominant CPU cost in our deployment.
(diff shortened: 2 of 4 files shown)


acockburn:
Hi there - thanks for the contribution. The PR is currently failing commit tests, could you fix please? Than we can get this merged.

jkempson:
Pushed a fix for the ruff isort failure (single-line import ordering tweak in the new test file). The new workflow runs are showing `action_required` and need maintainer approval to start. Thanks!

jsl12 (approved):

Issue #2532, opened 2026-01-25
Problem with self signed certificate and HASS connection


SvenKo:
### What happened?

Hi everyone, I'm using AppDaemon (4.5.7) and Home Assistant (2026.1.2) via Podman, which works great. However, I can only use AppDaemon up to version 4.5.7, as all subsequent versions display an SSL connection error. I get the same error when I use Docker.

All certificates I'm using are signed by my own CA. Even when I set ```cert_verify: False```, I still get this error. I've also entered my own CA's path ````cert_path: '/conf/cert/ca-cert.pem````, but that doesn't work either.

Do you have any idea what this could be, or is it perhaps a bug?

Thanks, Sven

### Version

4.5.13

### Installation type

Docker container

### Relevant log output

```sh
2026-01-25 07:48:19.003168 ERROR HASS: Cannot connect to host host.containers.internal:8123 ssl:True [SSLCertVerificationError: (1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain (_ssl.c:1010)')]
2026-01-25 07:48:19.003320 INFO HASS: Attempting reconnection in 5.0s
```

### Relevant code in the app or config file that caused the issue

```sh
appdaemon:
  latitude: ...
  longitude: ...
  elevation: 100
  time_zone: Europe/Berlin
  plugins:
    HASS:
      type: hass
      ha_url: https://host.containers.internal:8123
      token: ...
      cert_path: /conf/cert/ca-cert.pem
      cert_verify: False
    MQTT:
      type: mqtt
      namespace: mqtt
      client_id: "appdaemon"
      client_host: host.containers.internal
      client_port: 1883
      client_user: "mqttusr"
      client_password: "..."
      client_topics: NONE
http:
  url: https://appdaemon.mydomain.de:5050
  password: ...
  ssl_certificate: /conf/cert/appdaemon_mydomain_de-cert.pem
  ssl_key: /conf/cert/appdaemon_mydomain_de-key.pem
admin:
api:
hadashboard:
```

yraghu:
I am seeing this same issue as well. 
I revert back to version 4.5.12 and the problem is gone. 

It seems like version 4.5.13 doesn't properly use the cert_verify: False from the yaml configuration file.

yraghu:
Looking at the diff from 4.5.12 -> 4.5.13 yields this which is likely the problem...
```     def create_session(self) -> aiohttp.ClientSession:
[...]

PR #2663, opened 2026-08-08
Fix  HASS configuration `cert_verify` not working correctly

stablestud:
This commit fixes a regression that caused HASS config option `cert_verify` to not be applied correctly.


stablestud:
Fixes #2532

